PR #211, opened 2014-11-18
make it to work in non-bash compatible shells

tex:
bash scripts need bash


airblade:
vim-gitgutter now supports the fish shell (0ca15c8).

PR #219, opened 2014-12-18
Dealt with the case where the preview window is used by another plugin

lynnard:
I use fugitive a lot and there are times where I opened fugitive's preview window and then jumped to some part of the code trying to bring up the hulk previews for that particular part; where it fails because it's not able to reuse fugitive's preview window. I feel it's better to make it more generic like this so that the user doesn't have to close the preview window


airblade:
I've merged this in 25257c4.  Thanks again.

PR #233, opened 2015-03-03
Use a temporary file for buffer for realtime diffs

### LINKED ISSUE #232: Realtime diffs say entire file has changed if EOL differs between index and working tree, ignoring EOL conversion
When using vim-gitgutter on a system/repo that has CRLF conversion enabled (i.e. `core.autocrlf` is `true` or `input`), realtime diffs that run when the file hasn't been written to disk show every line as having been modified.  This is due to the fact that `git diff` does not perform EOL conversion on stdin, and `git show` never performs EOL conversion.  A workaround is to set `g:gitgutter_diff_args` to `'--ignore-space-at-eol'`, but this has the side effect of also making it not identify lines that have extra whitespace added at the end.  A better solution would be for vim-gitgutter to write the buffer to a temporary file and run `git diff --no-index blobfile tempbufferfile`.

elyscape:
`git diff` doesn't perform EOL conversion on stdin, causing it to mistakenly flag every line as having changed when the working tree uses a different EOL than the blobs. Writing the buffer to a temporary file and diffing against that avoids this issue.  This PR also deletes the temporary blob file created for realtime diffs.

Fixes #232.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -16,12 +16,19 @@ function! gitgutter#diff#run_diff(realtime, use_external_grep, lines_of_context)
   if a:realtime
     let blob_name = ':'.gitgutter#utility#shellescape(gitgutter#utility#file_relative_to_repo_root())
     let blob_file = tempname()
+    let buff_file = tempname()
+    let extension = gitgutter#utility#extension()
+    if !empty(extension)
+      let blob_file .= '.'.extension
+      let buff_file .= '.'.extension
+    endif
     let cmd .= 'git show '.blob_name.' > '.blob_file.' && '
+    execute('silent write '.buff_file)
   endif
 
   let cmd .= 'git diff --no-ext-diff --no-color -U'.a:lines_of_context.' '.g:gitgutter_diff_args.' -- '
   if a:realtime
-    let cmd .= blob_file.' - '
+    let cmd .= blob_file.' '.buff_file
   else
     let cmd .= gitgutter#utility#shellescape(gitgutter#utility#filename())
   endif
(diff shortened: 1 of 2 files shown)


elyscape:
This also removes `gitgutter#utility#buffer_contents()`, as it was only used when performing realtime diffs.

airblade:
Thanks for all this.  I didn't know that `git-diff` doesn't perform EOL conversion on stdin.

Why delete the temporary files after use?  It's neat and tidy but isn't it handled automatically by the OS?

I'm sure it works but I'd like to verify it myself.  Can you give me steps to reproduce the problem, so I can see this patch fixing it?  I'm on OS X.

airblade:
I tried this on lattesir/WindowsViaCPP which is linked from the [GitHub help on line endings](https://help.github.com/articles/dealing-with-line-endings/) as an example of a repo which enforces CRLF line endings.  It didn't work: the realtime diff flagged every line as changed.

When I execute `git show :Clean.bat > foo` manually, I can see that `Clean.bat` has CRLF but `foo` has LF.  Therefore diffing the buffer-in-a-temp-file, which has CRLF endings, against `foo` with its LF endings makes it look like every line has changed.

airblade:
Hmm, after cloning that repo `Clean.bat` had CRLF line endings as expected.  I played around with it, then `rm`'d it and checked it out again...whereupon it had LF line endings.  I didn't expect that.

elyscape:
> Why delete the temporary files after use? It's neat and tidy but isn't it handled automatically by the OS?

Vim's `tempfile()` function just returns the name of a temporary file that doesn't exist. The OS may eventually delete files in the temporary directory but depending on the system that may not be for a long time, so it's a good idea to clean up the files oneself.

elyscape:
> I tried this on lattesir/WindowsViaCPP which is linked from the GitHub help on line endings as an example of a repo which enforces CRLF line endings. It didn't work: the realtime diff flagged every line as changed.

Hmm.  Let me do some investigation.

elyscape:
I found the issue.  It happens because the tempfiles have no extensions.  I have squashed a fix into the second commit.

airblade:
I wouldn't have expected the lack of extensions to cause this; good catch.

I can't seem to see your fix though.

elyscape:
I squashed it into the second commit.  The easiest way to get it is to do a `git fetch` on it and then do `git reset --hard FETCH_HEAD`.

airblade:
I still can't seem to see it locally or on GitHub.  What was the specific change over your original commits?  Maybe it's there but I'm missing it.

elyscape:
Well that's weird.  It seems to have vanished somehow.  Let me redo it.

elyscape:
Okay, check again.

airblade:
Looks good, thanks!  Merged in 1fae335...d613f0f3.

elyscape:
Fantastic.  Thank you!

PR #235, opened 2015-03-03
Don't use --color if grep doesn't support it

### LINKED ISSUE #234: gitgutter no longer works on systems with GNU grep <2.5
1fae335 adds the `--color=never` flag to the external grep command.  Unfortunately, the `--color` flag was added to grep in 2.5, and earlier versions will fail:
```shellsession
$ grep --version
grep (GNU grep) 2.4.2

Copyright 1988, 1992-1999, 2000 Free Software Foundation, Inc.
This is free software; see the source for copying conditions. There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$ grep --color=always -e log CHANGELOG.md
grep: unrecognized option `--color=always'
Usage: grep [OPTION]... PATTERN [FILE]...
Try `grep --help' for more information.
```
Git for Windows ships with grep 2.4.2, by the way,

This manifests as gitgutter not identifying anything as having changed.

elyscape:
Not all versions of grep support the `--color` flag. This checks the output of `grep --help` when building the grep command and avoids using flags that aren't compatible with the version present.

Fixes #234.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -1,5 +1,12 @@
 let s:grep_available = executable('grep')
-let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep').' --color=never -e '.gitgutter#utility#shellescape('^@@ ')
+if s:grep_available
+  let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep')
+  let s:grep_help = gitgutter#utility#system('grep --help')
+  if s:grep_help =~# '--color'
+    let s:grep_command .= ' --color=never'
+  endif
+  let s:grep_command .= ' -e '.gitgutter#utility#shellescape('^@@ ')
+endif
 let s:hunk_re = '^@@ -\(\d\+\),\?\(\d*\) +\(\d\+\),\?\(\d*\) @@'
 
 


auscompgeek:
Hmm. Colours are disabled by default, so this only affects those that set `GREP_OPTIONS` anyway. Could we perhaps check `GREP_OPTIONS` to see if it contains `--color` instead?

airblade:
@auscompgeek This would also affect those with an alias along the lines of `alias grep='grep --color=always'`.  So I thinking checking `GREP_OPTIONS` isn't enough.

autoload/gitgutter/diff.vim
```diff
@@ -1,5 +1,12 @@
 let s:grep_available = executable('grep')
-let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep').' --color=never -e '.gitgutter#utility#shellescape('^@@ ')
+if s:grep_available
+  let s:grep_help = gitgutter#utility#system('grep --help')
+  let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep')
+  if s:help =~# '--color'
```
airblade: This should be `s:grep_help`.

elyscape: Good catch.  Fixed.

autoload/gitgutter/diff.vim
```diff
@@ -1,5 +1,12 @@
 let s:grep_available = executable('grep')
-let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep').' --color=never -e '.gitgutter#utility#shellescape('^@@ ')
+if s:grep_available
+  let s:grep_help = gitgutter#utility#system('grep --help')
+  let s:grep_command = ' | '.(g:gitgutter_escape_grep ? '\grep' : 'grep')
+  if s:grep_help =~# '--color'
```
airblade: One final tiny thing: please could you swap lines 3 and 4?  The grep command is the primary thing here so I think things are a little clearer when it's mentioned first...

elyscape: Makes sense.  Done.

airblade:
It's in d59ac03.  Thanks again!

PR #236, opened 2015-03-05
Echo message if no next/previous hunk is available

mantoni:
Gives some feedback when hunk navigation does nothing.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -39,10 +39,11 @@ function! gitgutter#hunk#next_hunk(count)
         let hunk_count += 1
         if hunk_count == a:count
           execute 'normal!' hunk[2] . 'G'
-          break
+          return
         endif
       endif
     endfor
+    echo 'No more hunks'
   endif
 endfunction
 
@@ -56,10 +57,11 @@ function! gitgutter#hunk#prev_hunk(count)
         if hunk_count == a:count
           let target = hunk[2] == 0 ? 1 : hunk[2]
           execute 'normal!' target . 'G'
-          break
+          return
         endif
       endif
     endfor
+    echo 'No previous hunks'
   endif
 endfunction
 


airblade:
Thanks for this.

The square bracket motions (at least the few I tried) don't give any feedback when you try to move too far, and I think of hunk navigation as square bracket motions.  In general I try to be consistent with vim's behaviour so I'm disinclined to make this change.

Another aspect is whether the cursor should move when you try to move further than possible.  The square bracket commands move to the furthest available location but the `f` command doesn't move the cursor at all.

The hunk navigation commands don't move the cursor when you try to move too far, which is consistent with `f` but not the square bracket motions.  By my own logic I should change this...

mantoni:
The only square bracket motions that I've tried is `:vimgrep` and then navigating the matches with `[q` and `]q`. They echo the file name on each navigation and `E553: No more items` with red background and white text on the last match. I think giving some feedback why a given operation does not work is helpful in any case and it does not come with any drawbacks. I'd argue that the other navigations should also adopt to the quick navigation list behavior.

airblade:
That makes sense.

Let's imagine there are two hunks above the current cursor position and the user presses `3[c`.  What do you think should happen?

- The cursor moves up 2 hunks; no feedback.
- The cursor moves up 2 hunks; feeds back it couldn't go further.
- The cursor doesn't move; feeds back it can't move 3 hunks.
- (The cursor doesn't move; no feedback – the current situation.)

I prefer the 1st option and I think you're advocating the 3rd – is that right?

mantoni:
Hm. Didn't think about this use case. I guess the first option works for me as well. The only thing I want is feedback if nothing happens at all.

airblade:
Your change is in 275d598 (and tweaked in 1be63d7).  In the end I decided to stick with the third option above and see how it goes.

mantoni:
Cool, thanks :+1:

PR #255, opened 2015-07-10
Fix corrupted characters in latin1 terminals

h3xx:
At the very least, this fix makes it so vim-gitgutter follows the same m.o. as [vim-airline](https://github.com/bling/vim-airline) and
[cream-showinvisibles](https://github.com/albfan/cream-showinvisibles) for showing Unicode.

Based on the settings of `&termencoding`, `&encoding` and
`has('gui_running')` we should decide whether to show UTF-8 characters.
These settings get set [and sometimes reset after your vimrc loads]
automagically based on:
- how Vim interprets what you set your TERM environment variable
- whether the value of LC_MESSAGES =~ /.utf8/
- whether the value of LC_CTYPE =~ /.utf8/
- evil spirits summoned through the value of TERMCAP
- the phase of the moon
- how many goats you've sacrificed to RMS in the last week

I don't understand it fully, but I tested it in 3 environments and this
fix works in all 3 of them.

### DIFF
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -38,11 +38,11 @@ call s:set('g:gitgutter_eager',                       1)
 call s:set('g:gitgutter_sign_added',                '+')
 call s:set('g:gitgutter_sign_modified',             '~')
 call s:set('g:gitgutter_sign_removed',              '_')
-try
+if has("gui_running") || &termencoding == "utf-8"
   call s:set('g:gitgutter_sign_removed_first_line', '‾')
-catch /E239/
-  let g:gitgutter_sign_removed_first_line = '_^'
-endtry
+else
+  call s:set('g:gitgutter_sign_removed_first_line', '_^')
+endif
 
 call s:set('g:gitgutter_sign_modified_removed',    '~_')
 call s:set('g:gitgutter_diff_args',                  '')


airblade:
What error do you get without this?  Is it one of Vim's numbered errors or something else?

Also what are the three environments you tested on?

airblade:
When I do `:set enc=latin1` and delete the first line of a git-tracked file, I still see the overbar sign correctly.

The overbar doesn't seem to have caused any problems for ages so I'm going to close this for now.  Happy to revisit if it's still problematic.

h3xx:
> When I do `:set enc=latin1` and delete the first line of a git-tracked file, I still see the overbar sign correctly.

Try `set termencoding=latin1` since that's the setting you need to set in order for vim to translate characters being sent to the terminal.

The only "error" feedback I get is how it looks in the terminal:

screenshot-2018-0214-1503

h3xx:
For comparison, this is how it looks with my changes:

screenshot-2018-0214-1540

airblade:
Ah, ok, I see what you mean.

Normally my terminal Vim has an `encoding` of `utf-8`, `termencoding` of an empty string, and `has("gui_running")` is `0` – and the overbar displays correctly.  But your patch would prevent the overbar being displayed for me.

h3xx:
Okay, I changed it to check the value of `&encoding` instead of `&termencoding`.

PR #256, opened 2015-07-11
Add support for fish shell

manuelmessner:
The `&&` and the `||` operators aren't available in fish.
The equivalents are `; and` and `; or`.
Also single parentheses are use for command substitution.
The fish equivalents are `begin` and `end`.
But they aren't needed here.


airblade:
Thank you very much.  Your patch is in 0ca15c8.

Issue #252, opened 2015-07-05
Show staged, but uncommited changes


somini:
Is it possible to show hunks that have been staged, but not yet commited?
Basically, the output of `git diff --cached`
I read the documentation but I couldn't find any option regarding this?

airblade:
This should do it: `let g:gitgutter_diff_args='cached'`.

airblade:
See also #249.

somini:
Great, thanks! That other issue solves half my "problem".

What I was really looking for was to keep it all in a single gutter. Meaning staged changes would show up in a different colour, for example.
Maybe it needs to run git 2 times, so it would slow things down considerably. How easy is to add extra symbols?

airblade:
To get both sets of changes you'd have to do everything twice which would slow things down.  I have found that people are quite sensitive to any lag this plugin causes.

You'd need to add a second set of signs, and keep track of a second set of hunks, and so on.  I suppose in principle it's straightforward, though fiddly.

It's not something I'm going to spend any time on but I'd be happy to try to help with any questions.

Issue #249, opened 2015-05-24
Showing cached changes


albfan:
Sometimes is useful to show what you have already staged, to unstage, modify or improve.

Adding this to ~/.vimrc is enough. See `-M` allow to detect renames showing real changes, nor successions of delete/add files

``` viml
let g:gitgutter_cached_enabled=0
function! GitGutterToggleCached()
  if g:gitgutter_cached_enabled 
    let g:gitgutter_diff_args=g:gitgutter_diff_args_old
  else
    let g:gitgutter_diff_args_old=g:gitgutter_diff_args
    let g:gitgutter_diff_args=g:gitgutter_diff_args." -M --cached"
  endif
  let g:gitgutter_cached_enabled=!g:gitgutter_cached_enabled
  GitGutterAll
endfunction

function! GitGutterIsCached()
  echo (g:gitgutter_cached_enabled ? "Enabled" : "Disabled")
endfunction

command! GitGutterIsCached :call GitGutterIsCached()
nnoremap <silent> <Leader>hc :call GitGutterToggleCached()<CR>
```

The missing enhancements are to show different symbols when cached is active:

```viml
call s:set('g:gitgutter_sign_added_cached',                'A')
call s:set('g:gitgutter_sign_modified_cached',             'M')
call s:set('g:gitgutter_sign_removed_cached',              'R')
call s:set('g:gitgutter_sign_removed_first_line_cached',   '^')
call s:set('g:gitgutter_sign_modified_removed_cached',    'MR')
```

I didn't create a PR, because I don't know where to put all these stuff. If you find it useful just let me know

airblade:
Thanks!  I'll be able to look at this properly when I return from holiday.

P.S.  Sorry to be brief; sent from my phone.

somini:
Yes, this almost works, changing the symbols is must, otherwise I get confused which mode is on.
To change the symbols at runtime, just run
`call gitgutter#highlight#define_signs()`
after [changing it](https://github.com/airblade/vim-gitgutter/issues/254#issuecomment-121214943).

albfan:
Great @somini! I would try to provide a PR based on that.

airblade:
@albfan Thanks!  I'd be happy to merge a pull request with this stuff.

[...]

PR #257, opened 2015-07-18
Provide a way to see staged changes

somini:
As referenced in #249 and #252.

This addition is disabled by default, per the `g:gitgutter_cached` variable. Well, unless it runs the `gitgutter#cached_enable` function is disabled anyway. How can I fix this?

The defaults are as follows:
- 'A' for staged additions
- 'M' for staged modification
- 'R' for staged removals
- '-M --cached' as the default diff arguments
  All this can be overridden just like the rest of the plugin.

It defines a new mapping '\hc' that toggles between the default
configuration and staged mode.

The option for running `git diff` twice and juggle all markers is left as an exercise to the reader.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /doc/tags
 /misc
 /test/*.actual
+/test/*.logvim
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -252,17 +252,6 @@ endfunction
 " Staged {{{
 function! gitgutter#staged_enable()
 
-  " Don't clobber the saved signs
-  if !g:gitgutter_staged
-    let g:gitgutter_sign_added_original = g:gitgutter_sign_added
-    let g:gitgutter_sign_modified_original = g:gitgutter_sign_modified
-    let g:gitgutter_sign_removed_original = g:gitgutter_sign_removed
-  endif
-
-  let g:gitgutter_sign_added = g:gitgutter_sign_staged_added
-  let g:gitgutter_sign_modified = g:gitgutter_sign_staged_modified
-  let g:gitgutter_sign_removed = g:gitgutter_sign_staged_removed
-
   let g:gitgutter_staged = 1
   call gitgutter#highlight#define_signs()
 
@@ -270,13 +259,6 @@ function! gitgutter#staged_enable()
 endfunction
 
 function! gitgutter#staged_disable()
-
-  " Signs
-  if g:gitgutter_staged
-    let g:gitgutter_sign_added = g:gitgutter_sign_added_original
-    let g:gitgutter_sign_modified = g:gitgutter_sign_modified_original
-    let g:gitgutter_sign_removed = g:gitgutter_sign_removed_original
-  endif
 
   let g:gitgutter_staged = 0
   call gitgutter#highlight#define_signs()
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -241,6 +241,10 @@ function! gitgutter#preview_hunk()
       setlocal noro modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
       execute "%delete_"
       call append(0, split(diff_for_hunk, "\n"))
+      " Delete the last, empty line
+      if empty(getline('$'))
+        execute '$delete_'
+      endif
 
       wincmd p
     endif
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -19,9 +19,30 @@ function! gitgutter#process_buffer(bufnr, realtime)
     endif
     try
       if !a:realtime || gitgutter#utility#has_fresh_changes()
-        let diff = gitgutter#diff#run_diff(a:realtime || gitgutter#utility#has_unsaved_changes(), 1)
-        call gitgutter#hunk#set_hunks(gitgutter#diff#parse_diff(diff))
-        let modified_lines = gitgutter#diff#process_hunks(gitgutter#hunk#hunks())
+        if g:gitgutter_staged
+          let diff_head  = gitgutter#diff#run_diff_revision(a:realtime || gitgutter#utility#has_unsaved_changes(), 1, 'HEAD')
+          let diff_index = gitgutter#diff#run_diff_revision(a:realtime || gitgutter#utility#has_unsaved_changes(), 1, '')
+          let hunks_head  = gitgutter#diff#parse_diff(diff_head)
+          let hunks_index = gitgutter#diff#parse_diff(diff_index)
+          " Remove the non-staged hunks
+          let current_hunks = filter(copy(hunks_head), 'index(hunks_index,v:val) == -1')
+
+          " TODO: Set the correct hunks
+          " call gitgutter#hunk#set_hunks()
+          let processed_head = gitgutter#diff#process_hunks(hunks_head)
+          let processed_index = gitgutter#diff#process_hunks(hunks_index)
+          let modified_lines = filter(copy(processed_head), 'index(processed_index,v:val) == -1')
+          if &verbose
+            echom 'P_H:'.string(processed_head)
+            echom 'P_I:'.string(processed_index)
+            echom 'RES:'.string(modified_lines)
+          endif
+        else
+          let diff = gitgutter#diff#run_diff(a:realtime || gitgutter#utility#has_unsaved_changes(), 1)
+          let current_hunks = gitgutter#diff#parse_diff(diff)
+          call gitgutter#hunk#set_hunks(current_hunks)
+          let modified_lines = gitgutter#diff#process_hunks(gitgutter#hunk#hunks())
+        endif
 
         if len(modified_lines) > g:gitgutter_max_signs
           call gitgutter#utility#warn_once('exceeded maximum number of signs (configured by g:gitgutter_max_signs).', 'max_signs')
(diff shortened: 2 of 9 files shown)


albfan:
For me this work nice! I have removed my quick&dirty solution.

I'm interesting too on show all markers to confront a complex situation: Move changes on a staged file

http://stackoverflow.com/questions/9500645/editing-git-patch-gives-your-edited-hunk-does-not-apply

Solve that needs some logic that a vim plugin should provide. My thoughs are to empower GitGutterPreviewHunk to allow move changes between worktree and index.

somini:
I've rebased this to the last master.

airblade:
Thanks @somini and @albfan.  I'm really interested in merging this but I want to make sure I understand it all first, and unfortunately I'm rather busy at the moment.  I'll get to it as soon as I can...

albfan:
Let's define some test about this functionality. That would be the best way to understand what's happening underneath and asure a intuitive behaviour.

I'm use to [vader](https://github.com/junegunn/vader.vim) to test vim scripts  and [sharness](https://github.com/mlafeldt/sharness) due to it provides a sandbox for test out of the box. Here you use [test/helper.vim](https://github.com/airblade/vim-gitgutter/blob/master/test/helper.vim) but it should be easy to adapt any existing test to check staged changes

airblade:
I test this plugin [like this](https://github.com/airblade/vim-gitgutter/blob/master/test/README.markdown).  I'm not going to switch test frameworks as part of this pull request...

somini:
After using this functionality some more, I found a bug that might require an overhaul, it's not as simple as switching a git flag.

The location of the changes are only correct if the line numbers don't change. If the unstaged changes include new lines, the gutter has offsets that match those new lines. It's because when running `git diff --staged` compares the staged with the last commit, when what we want is to compare the staged with the current version of the file.

I'll hunt around the git commands, I'm sure there's an obscure flag that does what we want.

airblade:
Well, yes :)  To diff staged changes, and potentially unstage them, you have to diff the index against the repo instead of the working tree against the index.  Neither the index nor the repo are what's in your buffer in Vim.

albfan:
@somini git diff --no-index is probably what you're looking for. 

example

    $ mkdir repo
    $ cd repo
    $ git init
    $ cat >file1 <<EOF
    foo
    baz
    bar
EOF
    $ git add .
    $ git commit -m "first file"
    $ cat >file1 <<EOF
    foo
    baz
    hoge
    fuga
    bar
    EOF
    $ git add .
    $ cat >file1 <<EOF
    foo
    baz
    bar 
    EOF

From here

unstaged file

    $ cat file1

file cached:

    $ git show :file1

base file:

    $ file show HEAD:file1

redir to **/tmp/file1_cached** or **/tmp/file1_unstaged** and use `git diff --no-index`

    $ git diff --no-index file1 /tmp/file1_cached

Hope it helps

somini:
@albfan Thanks, but that's not the problem. Here's an image from SO that helps to visualize the problem.
GitDiff
What we want is none of those.

The only solution I can see is storing the amount of lines added or removed and adjust the following markers, but that leads to running git diff twice, and introducing a lot of complexity.

albfan:
> what we want is to compare the staged with the current version of the file.

I'm pretty sure:

    $ cd repo
    $ git show :file1 > /tmp/file1.cached
    $ git diff --no-index file1 /tmp/file1.cached

compares what you want. Consider create a dummy repo with the problem. Zip it on a branch on your fork or reproduce it with a script from bash command line (as I show you) I guess that would be the way to check this.

Maybe this is related with applying edited chunk (unstaged changes are most like an edited destination instead of edited change)

http://stackoverflow.com/a/3270837/848072

somini:
I had a go at this before, and I did verify that @albfan 's solution didn't work, but today I made sure. :D
Here's a zip with a repo containing a single file, `a`. Using `\hc` on the buffer, you can see the 3 extra lines at the bottom are shifted 3 lines upwards.

http://b.1339.cf/wzyionu.zip

The target diff is the following:

    6,8d2
    < 22
    < 23
    < 24


After fiddling around, I found a "solution". It involves 2 diffs and removing the common chunks. Probably too hard to make in VimL, but as a awk script or something it seems feasible.
I included the diffs in the zip.
The objective can be made by taking `diff_a-head`(working directory > HEAD|git diff HEAD) and then removing the common chunks from `diff_a-cached`(working directory > index|git diff). Do notice this is different from the output of `git diff --cached`, the locations are different.

somini:
Ironically, only after this I noticed that "removing common chunks" means diffing (or something similar), so here's a command that does exactly what I want:

    $ comm -13 --nocheck-order diff_a-cached diff_a-head

Yes, `comm` is a common command, it's in coreutils.

somini:
I think this does it:

    $ comm -13 --nocheck-order <(diff a <(git show :a)) <(diff a <(git show HEAD:a))

Now the only thing to do is make gitgutter run this, and make it understand the path it munges. It also requires bash trickery and coreutils, so Windows support seems flaky. We could have a fallback with temporary files...

albfan:
We are just a test case away from solution. 

I suggest to create a branch with a test failing, maybe once proposed chances for others to solve arise. 

Needs for test:

- [ ] Reproduce @somini zip repo
- [ ] Create a theorical expected signs dump.

A good move will be to connect vim-gutter to a CI -maybe travis CI?- and instruct to pass test, that way any proposal will be automatically verified against test and adding new test will check the strongness for solution. Here is a naive example:

- https://travis-ci.org/albfan/git-ignore
- https://github.com/albfan/git-ignore/blob/master/.travis.yml

airblade:
> `comm -13 --nocheck-order <(diff a <(git show :a)) <(diff a <(git show HEAD:a))`

At one stage vim-gitgutter used process substitution but I had to get rid of it because it broke things for so many people.  Everyone using zsh, or Fish, or Windows will have problems.

Similarly using `comm` is out of the question because many people won't have it.

If you want to adjust line offsets to account for unsaved changes, have a look at how the hunk staging/reverting code does it.

vim-gitgutter's current diffing process works for everyone as far as I know, and it took a lot of trial and error to get there.  I learned that you really do have to keep things simple and fast.

albfan:
I think @somini comments are only the quick and dirty. The process substitution can be replace essily with redir.

Whatever comm does can be replace with vimL, but I agreed that this would be valid only if it keep plugin fast and intuitive.

somini:
Yes, @albfan is correct, I was merely creating a proof of concept. Now that I have a target to strive for, I can start developing it.
I'll start by reading the code that deals with the hunks, and see if I can work something out.
I'll also write a failing test using the example repo I posted.

somini:
Rebased this into the latest master.
Still haven't got time to get this started...

somini:
Well, I have preliminary results to share.

I managed to implement (monkey-patched, more like it) this in my fork, but it does require 2 diffs, so the performance might not be acceptable on certain repositories.
What I did was compare the hunks by diffing against `HEAD` and the index. The hunks that correspond to the staged changes exist on the `HEAD` diff, but not on the index diff.
The I tested on my special repo and everything looks OK.

To test this, `call GitGutter_StagedChanges()` on a file to see the staged changes. This is not my final PR, I'll integrate it properly with the rest of the plugin, but it was useful to get my bearings around the code base.
I'll clean this up and have a mergeable version of this next week.

somini:
I've squashed and rebased this into a more usable version.
The previous naïve version failed when some commits added or deleted lines. I solved that by using the function `gitgutter#diff#process_hunks` and "diffing" on the output.

The results are correct, but with that I can't have a hunks object, so hunk movement doesn't work. That function does the magic of adjusting the line numbers when they change.
Another feature that doesn't work is the hunk stage/reset/preview. I've stubbed them out for now, but I would like for them to actually do stuff.

I'll see if I can write a test case for this, it would make it much easier to change without fear of regressions.

somini:
Here's a test case for the feature so far.

Commit 55d3e56551b4d31f1c89c9e87ac612768a544233 is independent, can be cherry-picked into master.

somini:
I've rebased this into the latest master, there was a conflicting line and it's better to solve conflicts early.

I've been using this for a while now and can't find any bugs so far.

somini:
Commit https://github.com/somini/vim-gitgutter/commit/e648217c28a63ddf8b08546942a0462d6bf6d395 is also independent, can be cherry-picked into master. Just adjust the message.

The other commits add a much cleaner way to handle different symbols in different modes and preliminary support for `\hr`.
That mappings resets a hunk, that is unstages it. It works, but since it depends on the backend hunks being correctly defined, there's still an offset from the real place.

I detected other issue, we should also patch vim-airline to use the different symbols in different modes, but that's an easy thing, only when this is in the brink of merge.

somini:
Ping @airblade 
I see you've been updating the plugin to work with nvim jobs, so this doesn't merge cleanly now.

somini:
Can't believe it has been 4 months.

This is mostly working, the only thing I really wanted to do is hunk movement. See https://github.com/airblade/vim-gitgutter/pull/257#issuecomment-173040323
I'll see if I can find some time to rebase this to the lastest master, there have been many changes.

robinfehr:
is this already working on the master? quickly tried to get it working according to the comments up here. if yes, it would be nice to have a quick installation guide 💃 

thank you.

somini:
Last time I hacked on this, the only missing part was hunk movement in staged mode.
It does two diffs,so for large repos it might be two slow. 
Master has moved a lot, so this would need some more work to merge cleanly.

somini:
Can't believe it has been over 1 year!

I rebased this to the current repo, but since it's so different, I'm opening a different PR #452.

PR #258, opened 2015-07-24
Wipeout temporary buffer created when writing file to diff

RedBeard0531:
This avoids creating a ton of dead buffers that slow down plugins that
iterate over all buffers, such as YouCompleteMe.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -75,6 +75,7 @@ function! gitgutter#diff#run_diff(realtime, use_external_grep)
   if a:realtime
     call delete(blob_file)
     call delete(buff_file)
+    execute 'keepalt silent! bwipeout' buff_file
   endif
 
   if gitgutter#utility#shell_error()


airblade:
Many thanks for this.  I usually use `:ls` or similar to list the buffers so I never noticed those temporary ones were there, unlisted.

Tranquility:
Not sure why, but I don't have an unlisted buffer without this PR. That means now with this PR whenever this line gets executed I get an ugly error message that some file in `/tmp/*` does not exist.
I fixed it like this: https://github.com/Tranquility/vim-gitgutter/commit/7e12ec8161161221c9bb3e819bbd90a2bebe06f7 would you apply this patch?

But it would probably better to find the real cause. Not sure what's different on my system.

airblade:
@Tranquility I'm happy to add the `!` but it would be nice to find real cause.  What version of vim are you using and is your vimrc online?

Tranquility:
I disabled all plugins except for gitgutter, removed everything from my vimrc and downgraded vim to 7.4.273 (which was the lowest I could get with the Gentoo linux package manager). But I don't get any buffers in `:ls!`.

airblade:
@Tranquility Did you also comment out the line in question which wipes out the buffers?  If so, it's surprising that you don't have any unlisted buffers after vim-gitgutter has made some "realtime" updates.

In the meantime I have added that `!` in 62cfae4.

Tranquility:
> @Tranquility Did you also comment out the line in question which wipes out the buffers?

Yes I did. I have no idea how the explain the different behavior. Thanks for adding the `!`.

If you want me to try anything else, please tell me.

PR #261, opened 2015-08-07
Make sign line filter expression more specific

andrewmwhite:
With previous expression, setting verbose>=15 
(which prints each line before it’s executed)
results in an exception because the “signs”
variable now includes the text of line 77, thus 
passing the filter. When the “name” variable 
assignment is then attempted, “components[2]”
contains no ‘=‘ signs and hence trying to access 
the second element of “split(components[2], ‘=‘)” 
results in an out-of-bounds exception (E684).

### DIFF
--- a/autoload/gitgutter/sign.vim
+++ b/autoload/gitgutter/sign.vim
@@ -77,7 +77,7 @@ function! gitgutter#sign#find_current_signs()
     silent execute "sign place buffer=" . bufnr
   redir END
 
-  for sign_line in filter(split(signs, '\n'), 'v:val =~# "="')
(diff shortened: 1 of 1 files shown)


airblade:
Thanks for reporting this.

I don't want to hardcode `line`, `id`, and `name` into the regexp because those only apply to the english locale.

A recent change (a30ee7f) ignores the first two lines of `signs` before filtering the rest.  I would have thought that would fix this problem but it doesn't.

I tried to bypass the problem by using `10verbose redir => signs` but that didn't seem to work.

Are you able to update your vim-gitgutter and post the output of `signs` at verbose 15?  I can't seem to get to it...

airblade:
I'm not too worried about this breaking with a non-standard verbosity.

PR #264, opened 2015-08-25
Specify help files correctly

somini:
This makes so that editing helpfiles directly triggers the gutters,
while keeping the default behaviour of editing help buffers (opened
with `:help stuff`) doesn't.


airblade:
Thanks!  It's in 1067294.

Issue #274, opened 2015-10-17
Error on file switching


damionjn:
Receive the following after working for a bit, having multiple buffers open then switching files with https://github.com/rking/ag.vim fuzzy finder:

```
Error detected white processing function gitgutter#process_buffer..gitgutter#diff#run_diff:
line 71:
E89: No write since last change for buffer 33
```

This causes a segmentation fault and crashes vim every time.

NVIM 0.0.0-alpha+201507262350
Using `nvim -p` command

After pressing enter to choose the file to edit:
![screen shot 2015-10-17 at 1 33 40 pm](https://cloud.githubusercontent.com/assets/7221494/10560979/de62ad5a-74d8-11e5-90c4-58c3b36f4ab8.png)

airblade:
It looks like vim-gitgutter is trying to process some "fake" buffer used by ag.vim.  ag.vim writes its search results in a split window: perhaps that's the one vim-gitgutter is having trouble with.

I think a workaround for the symptom might be to use `silent!` instead of `silent` [here](https://github.com/airblade/vim-gitgutter/blob/a30ee7f23e93c827b4857d337f04c1ddf3c0c16e/autoload/gitgutter/diff.vim#L85).  It would be better to address the cause, i.e. to stop vim-gitgutter processing ag.vim's non-file buffer.

Having said all that, I don't understand why the problem only manifests after a while, not the first time you use ag.vim.

damionjn:
I've been away doing some work on an iOS project, but had some time yesterday to take a look at this. It didn't crash a single time. I'm unsure if it's because I was doing small commits (only activating gitgutter on 3-4 files at a time), but I'll go ahead and say #275 has fixed the issue. Thanks!

airblade:
Thanks for the feedback.

So what version of vim-gitgutter are you running?  Is it the latest but with the patch in #275 applied?  I'm wondering whether you've got 62cfae4...

damionjn:
airblade... I have no idea why, but it seems to work fine now.

I just spent 20 minutes opening files specifically with ag.vim and adding new lines for vim-gitgutter to mark. I did *not* have the latest- so I installed and re-tested editing each file in my project and this version seems to work also.

airblade:
Thanks for testing it out again.  I'll assume everything's fine now unless I hear from you to the contrary.

PR #275, opened 2015-10-18
Make tracked file check more failure resistant

Tranquility:
Sometimes the gitgutter_tracked variable was not set which lead to an
ugly error message. This happened especially with untracked files in
neovim.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -17,7 +17,7 @@ function! gitgutter#diff#run_diff(realtime, use_external_grep)
   let cmd = s:fish ? '' : '('
 
   let bufnr = gitgutter#utility#bufnr()
-  let tracked = getbufvar(bufnr, 'gitgutter_tracked')  " i.e. tracked by git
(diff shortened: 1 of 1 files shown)


Tranquility:
This should fix #274

Tranquility:
Doesn't fix all problems after all. I will open again when I found the problem.

airblade:
I'm not sure why the change makes any difference.

From Vim's (and Neovim's) docs for `getbufvar({expr}, {varname} [, {def}])`:

> When the buffer or variable doesn't exist {def} or an empty string is returned, there is no error message.

So if `gitgutter_tracked` has not been set, we should get an empty string back without error message.

I do agree that it's inelegant to have `tracked` be either an empty string or the integer `1`.  It would be better to do provide a default of `0`:

```viml
let tracked = getbufvar(bufnr, 'gitgutter_tracked', 0)
```

Unfortunately a default value for `getbufvar()` was only introduced in [Vim v7.3.831](https://github.com/vim/vim/commit/63dbda1caa88042c88d46ad139c99def2e8b80ca) and I'd like to support Vim 7.3.  I suppose I could introduce my own polyfill method...

PR #280, opened 2015-11-07
Allow GitGutter* commands to be followed by other commands

knownasilya:
See the related discussion here: https://github.com/scrooloose/syntastic/issues/1593


airblade:
Thank you!

PR #291, opened 2015-12-14
Use git diff-files instead of git diff

projedi:
This improves performance on huge repositories. I didn't investigate what precisely is going on but it *seems* that `git diff` sometimes (e.g. when reverting all the changes in a file) causes index update which freezes vim. Reproducible by running in shell: `touch somefile && git diff -- somefile`. Running `git diff-files` instead avoids the freeze.

I'm not aware of any negative implications of the change: used it for a couple of month without spotting buggy behaviour.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -49,7 +49,7 @@ function! gitgutter#diff#run_diff(realtime, use_external_grep)
     call setpos("']", op_mark_end)
   endif
 
-  let cmd .= 'git diff --no-ext-diff --no-color -U0 '.g:gitgutter_diff_args.' -- '
+  let cmd .= 'git -c "diff.autorefreshindex=0" diff --no-ext-diff --no-color -U0 '.g:gitgutter_diff_args.' -- '
   if a:realtime
     let cmd .= blob_file.' '.buff_file
   else


projedi:
Just reproduced on a shallow clone of [chromium](https://chromium.googlesource.com/chromium/src.git) so here are some numbers:

Without changing file:
```
$ touch DEPS && time git diff -- DEPS

real	0m5.573s
user	0m0.478s
sys	0m4.927s
$ touch DEPS && time git diff-files -p -- DEPS

real	0m0.248s
user	0m0.111s
sys	0m0.030s
$ touch DEPS && time git diff-files -p -- DEPS

real	0m0.194s
user	0m0.156s
sys	0m0.038s
$ touch DEPS && time git diff -- DEPS

real	0m6.561s
user	0m0.466s
sys	0m5.803s
```

After changing file:

```
$ touch DEPS && time git diff -- DEPS
diff --git a/DEPS b/DEPS
index fe38648..abc7b5d 100644
--- a/DEPS
+++ b/DEPS
@@ -1,3 +1,4 @@
+# asdf
 # This file is used to manage the dependencies of the Chromium src repo. It is
 # used by gclient to determine what version of each dependency to check out, and
 # where.

real	0m0.139s
user	0m0.111s
sys	0m0.031s
$ touch DEPS && time git diff-files -p -- DEPS
diff --git a/DEPS b/DEPS
index fe38648..abc7b5d 100644
--- a/DEPS
+++ b/DEPS
@@ -1,3 +1,4 @@
+# asdf
 # This file is used to manage the dependencies of the Chromium src repo. It is
 # used by gclient to determine what version of each dependency to check out, and
 # where.

real	0m0.138s
user	0m0.116s
sys	0m0.032s
$ touch DEPS && time git diff-files -p -- DEPS
diff --git a/DEPS b/DEPS
index fe38648..abc7b5d 100644
--- a/DEPS
+++ b/DEPS
@@ -1,3 +1,4 @@
+# asdf
 # This file is used to manage the dependencies of the Chromium src repo. It is
 # used by gclient to determine what version of each dependency to check out, and
 # where.

real	0m0.141s
user	0m0.113s
sys	0m0.031s
$ touch DEPS && time git diff -- DEPS
diff --git a/DEPS b/DEPS
index fe38648..abc7b5d 100644
--- a/DEPS
+++ b/DEPS
@@ -1,3 +1,4 @@
+# asdf
 # This file is used to manage the dependencies of the Chromium src repo. It is
 # used by gclient to determine what version of each dependency to check out, and
 # where.

real	0m0.142s
user	0m0.112s
sys	0m0.033s
```

So, on a changed file it takes the same time either way, but on an unchanged file `git diff` takes 5 seconds while `git diff-files` takes 200 milliseconds.

airblade:
Thanks for the benchmarks.

I wonder why `git-diff` is 25 times slower than `git-diff-files` on an unchanged file?  It sounds like something that should be reported to the `git` project.

I couldn't find any information about `git-diff` versus `git-diff-files`.  Can you point to any?  While your benchmarks speak for themselves I'm reluctant to make the change without knowing a bit more.

airblade:
According to [this thread from 2012](http://git.661346.n2.nabble.com/Git-diff-file-bug-td7568132.html), `git-diff-files` doesn't update the index before diffing whereas `git-diff` does; the former is regarded as plumbing and the latter as porcelain.

Presumably if you run this benchmark it'll be similar to the `git diff` case:

```
touch DEPS && time git update-index --refresh && time git diff-files -p -- DEPS
```

Having said that I don't understand why it takes `git diff` so much longer to process an unchanged file than a changed one.

projedi:
Yeah, running `touch DEPS && time git update-index --refresh && time git diff-files -p -- DEPS` yields similar slowness.

Also I've found this change https://github.com/git/git/commit/aecbf914c43ab76e055fa8a25b12456ae74d91bc There on line 373 of builtin-diff.c there's a check that fires for an unchanged file but doesn't for a changed one and so index isn't forced to refresh. And indeed by turning `diff.autorefreshindex` off, performance on an unchaged file is adequate:
```
$ touch DEPS && time git diff -- DEPS

real	0m4.436s
user	0m0.412s
sys	0m3.368s
$ touch DEPS && time git -c 'diff.autorefreshindex=0' diff -- DEPS

real	0m0.268s
user	0m0.114s
sys	0m0.033s
```

Also tried using this in gitgutter and didn't experience freezes when reverting changes in a file. If I understand rightly, `diff.autorefreshindex=1` is used to automatically update stat information in the index, which does not seem a needed step in the plugin. Will update PR shortly.

Or maybe it makes more sense to make it an option, what do you think?

airblade:
Sorry for the delayed response – happy new year ;)

I think this plugin should only care about content differences, not stat-only changes.  So this seems like a promising patch.

I'm still a little hazy on what `diff.autorefreshindex` does.  I can see that when it's off it skips updating the stat information in the index for stat-only changes.

However the docs say that when it's on, `git-diff` does not consider a stat-only change as changed.  Does that mean that when it's off a stat-only change is considered a change?  We don't want that.

projedi:
Happy New Year to you too!

Yeah, by not updating the index, stat-only change is treated as a change, and you can see it with --raw. However, the diff patch output will still be empty

airblade:
Ok, that makes sense.  (Oddly, though, on my OS X the last of your reproduction steps didn't show any output; it did on Ubuntu.  Any ideas why OS X is different?)

Your patch is in 4510e9b.  Thanks very much!

projedi:
Cool, thanks!

That's weird - I've tested these steps on OS X with git versions 2.3.2, 2.5.3; on Linux with git 2.7.0; and on Windows with git 1.9.5. All behave the same and show a change with diff —raw. Could it be that on OS X touch didn't update mtime for some reason?

airblade:
I've tried with git 2.6.4 and 2.7.0 on OS X 10.11.2.  `touch` updates the mtime as verified by `stat README` and `ls -l README`.  Strange...

projedi:
I've also just tried with OS X 10.11.2 (the other one being 10.10.smth) with git 2.5.4 and it still works.

airblade:
So the problem isn't with OS X 10.11.2, and I doubt it's with git 2.6.4 / 2.7.0.  Therefore it's probably a problem with my git configuration.  However I moved all my `~/.git*` configs out of the way and nothing changed.  Very odd.

Oh well, I'll just assume it's a weirdness on my own system.  Thanks for looking into it!

airblade:
Progress...if I edit the repo's git config file, i.e. `./.git/config`, and add the following:

```
[diff]
  autorefreshindex = 0
```

Then I get the stat-change output – regardless of whether or not I use `-c 'diff.autorefreshindex=0`.

```
$ git init
$ echo 123 > README
$ git add .
$ git commit -m 'Initial'
[master (root-commit) f38a08f] Initial
 1 file changed, 1 insertion(+)
 create mode 100644 README
$ touch README
$ git -c 'diff.autorefreshindex=0' diff -- README
$ git -c 'diff.autorefreshindex=0' diff --raw -- README
$ echo -e "[diff]  autorefreshindex = 0" >> .git/config
$ touch README
$ git -c 'diff.autoRefreshIndex=0' diff --raw -- README
:100644 100644 190a180... 0000000... M  README
$ git -c 'diff.autoRefreshIndex=0' diff -- README
$ git diff --raw -- README
:100644 100644 190a180... 0000000... M  README
$ git diff -- README
```

PR #293, opened 2015-12-22
allow setting global g:gitgutter_notify_toggle to get notification on toggle

eyalfir:


airblade:
Thanks, but surely you just know whether you're turning the plugin on or off?

eyalfir:
My scenario is:
1. I have neomake (syntactic) plugin on to the gutter is always visible. So the fact the gutter is visible doesn't imply whether or not gitgutter is on/off.
2. I have a mapping for :GitGutterToggle so I'm not using GitGutterEnable nor GitGutterDisable directly.

Often I find myself looking at a piece of code, and to understand whether or not it's a diff I have to toggle gitgutter (since I'm not sure whether I have it on or off at that moment).

eyalfir:
My scenario is:
1. I have neomake (syntactic) plugin on. So the fact the gutter is visible doesn't imply whether or not gitgutter is on/off.
2. I have a mapping for :GitGutterToggle so I'm not using GitGutterEnable nor GitGutterDisable directly.

Often I find myself looking at a piece of code, and to understand whether or not it's a diff I have to toggle gitgutter (since I'm not sure whether I have it on or off at that moment).

airblade:
That makes sense.  However I think there's an easier way to find out whether gitgutter is on or off than toggling it once or even twice: show `g:gitgutter_enabled` in your status line; then you don't need to press any keys.

With bling/airline it could look like:

```viml
let g:airline_section_y = "GitGutter %{g:gitgutter_enabled ? 'on' : 'off'}"
```

eyalfir:
Indeed easier. Thanks.

Issue #260, opened 2015-07-31
How to make GitGutter{Prev,Next}Hunk to wrap around?


claudiodangelis:
Is it possible to make these two commands wrap around on reaching the beginning/ending of the file?

Thanks!

airblade:
There was a discussion on a similar topic in #236.

I suppose I could add a configuration setting which would allow/prevent those commands wrapping.

Just need a bit of spare time...!

claudiodangelis:
Awesome, thank you.

airblade:
I don't want to build this into the plugin: it's big enough already.  You can achieve what you want with this in your vimrc:

```viml
function! PrevHunkWrap(count)
  for i in range(1, a:count)
    let line = line('.')
    silent GitGutterPrevHunk
    if line('.') == line
      normal G
      silent GitGutterPrevHunk
    endif
  endfor
endfunction

function! NextHunkWrap(count)
  for i in range(1, a:count)
    let line = line('.')
    silent GitGutterNextHunk
    if line('.') == line
      normal 1G
      silent GitGutterNextHunk
    endif
  endfor
endfunction

command -count=1 Phw call PrevHunkWrap(<count>)
command -count=1 Nhw call NextHunkWrap(<count>)
```

PR #295, opened 2016-01-10
make it possible to wrap around with GitGutter{Prev,Next}Hunk

spacewander:
For issue https://github.com/airblade/vim-gitgutter/issues/260
Add a new option `g:gitgutter_circular_hunks` to configure if GitGutter{Prev,Next}Hunk will wrap around.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -34,16 +34,23 @@ function! gitgutter#hunk#next_hunk(count)
   if gitgutter#utility#is_active()
     let current_line = line('.')
     let hunk_count = 0
-    for hunk in s:hunks
-      if hunk[2] > current_line
-        let hunk_count += 1
-        if hunk_count == a:count
-          execute 'normal!' hunk[2] . 'G'
-          return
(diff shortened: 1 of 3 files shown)


airblade:
Thanks, I hope to take a look soon.

brendanmaguire:
Any chance of this getting merged? Would be nice to have.

airblade:
I don't want to build this into the plugin: it's big enough already.  You can achieve what you want with this:

```viml
function! PrevHunkWrap(count)
  for i in range(1, a:count)
    let line = line('.')
    silent GitGutterPrevHunk
    if line('.') == line
      normal G
      silent GitGutterPrevHunk
    endif
  endfor
endfunction

function! NextHunkWrap(count)
  for i in range(1, a:count)
    let line = line('.')
    silent GitGutterNextHunk
    if line('.') == line
      normal 1G
      silent GitGutterNextHunk
    endif
  endfor
endfunction

command -count=1 Phw call PrevHunkWrap(<count>)
command -count=1 Nhw call NextHunkWrap(<count>)
```

PR #329, opened 2016-04-25
Add a call to GitGutterProcessBufferHook, a user-defined hook

jdelkins:
This hook allows for easier integration with other user configuration,
including statuslines, etc.

One specific use case is integration with [lightline](https://github.com/itchyny/lightline.vim), when using information from GitGutterGetHunkSummary in a lightline ```component_expand``` segment.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -30,6 +30,9 @@ function! gitgutter#process_buffer(bufnr, realtime)
   else
     call gitgutter#hunk#reset()
   endif
+  if exists("*GitGutterProcessBufferHook")
+    call GitGutterProcessBufferHook(a:bufnr)
+  endif
 endfunction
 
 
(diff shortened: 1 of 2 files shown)


airblade:
Thanks for this.

An alternative approach would be for the plugin to fire a user autocommand.  I wonder which is more consistent with the way vim does things?

jdelkins:
Not sure. A user autocmd sounds like a more complete solution, but I was just trying to keep it simple. FWIW [Syntastic takes this approach](https://github.com/scrooloose/syntastic/blob/master/plugin/syntastic.vim#L345) as well. I'd be happy either way!

jdelkins:
I changed to a user autocmd approach with the 2nd commit. I think this is probably better overall. Also added documentation. Feel free to cherry pick whichever approach you prefer. Thanks!

airblade:
Thanks, I prefer the autocmd approach; it feel more decoupled.

Anyway, how about renaming it to just `GitGutter`?

jdelkins:
Done!

airblade:
I was about to merge this when I had a thought: don't we only want to fire the autocmd when `gitgutter#utility#is_active()` is truthy?  I.e. move the autocmd to the last line of that block, just above the `else`?

jdelkins:
I thought it over and I think you are correct.

airblade:
Many thanks!  It's in 27f99c4.

Issue #314, opened 2016-02-29
Signs "bleed through" to terminal buffer.


raichoo:
Hi, I'm using a recent version of neovim `NVIM v0.1.3-269-g9c4b34b` and I'm having the following effect.

When switching to a terminal buffer the signs of an "underlying" buffer bleed through to the terminal and it ends up looking like this:

![screenshot from 2016-02-29 10-49-11](https://cloud.githubusercontent.com/assets/151506/13391090/ad810c36-ded2-11e5-8cfe-e6a61c224257.png)

These signs are coming from another buffer:

![screenshot from 2016-02-29 10-49-29](https://cloud.githubusercontent.com/assets/151506/13391105/c0ede546-ded2-11e5-81c7-a928bd34b185.png)

Other non-terminal buffers are not affected and display their signs correctly.

Kind regards,
raichoo

airblade:
I have no idea what's causing this, unfortunately.  Do you get the same effect when placing a sign manually?  E.g.:

```viml
:sign define name=foo text=>>
:sign place 42 line=1 name=foo buffer=1
```

(Replace the buffer number with the appropriate value.)

raichoo:
The sign I put there myself (using your commands) does not "bleed" through. The gitgutter signs however do.

airblade:
How odd.

The only difference between the manual sign and gitgutter's signs are the latter's have foreground and background colours set.

If you edit the [highlight](https://github.com/airblade/vim-gitgutter/blob/master/autoload/gitgutter/highlight.vim) file to get rid of the `guibg` and `ctermbg` (lines 14-16), does that make a difference?

raichoo:
I've commented these lines out, same effect. I've noticed that some redraw event need to be triggered for this to happen. Switching desktops back and forth on Gnome3 triggers that event instantly, otherwise I have to wait for a while for this to happen (maybe the update timeout waiting in the background?).
I'm making wild guesses here, but I'm trying to provide as much information as possible.

airblade:
I appreciate that.  Like I said originally, I have no idea what's going on.  Maybe someone on the neovim end who knows the terminal buffer implementation might be able to shed some light?

raichoo:
I don't know to be honest. It looks like this issue is related to gitgutter, since I've only encountered this since I've started using it. This certainly won't stop me from using it in the future, it's just somewhat annoying. Anyway, maybe the terminal works in a way that needs some special handling, but I'm afraid I cannot answer that.

[...]

PR #330, opened 2016-04-29
Apply signs to correct buffer when using Neovim async

tweekmonster:
When using async, the buffer might change before job completion and apply signs to the wrong buffer.  This seems to happen most when splits that are created by plugins appear/disappear (fugitive, fzf).  This also appears to address signs not appearing when Neovim is started with a saved session, and may address #314 

I'm not sure if this is an issue when using mainstream Vim's async.

### DIFF
--- a/autoload/gitgutter/async.vim
+++ b/autoload/gitgutter/async.vim
@@ -10,6 +10,7 @@ endfunction
 function! gitgutter#async#execute(cmd)
   if has('nvim')
     let job_id = jobstart([&shell, &shellcmdflag, a:cmd], {
+          \ 'buffer': gitgutter#utility#bufnr(),
           \ 'on_stdout': function('gitgutter#async#handle_diff_job_nvim'),
           \ 'on_stderr': function('gitgutter#async#handle_diff_job_nvim'),
           \ 'on_exit':   function('gitgutter#async#handle_diff_job_nvim')
@@ -38,7 +39,10 @@ endfunction
 
 
 function! gitgutter#async#handle_diff_job_nvim(job_id, data, event)
-  call gitgutter#debug#log('job_id: '.a:job_id.', event: '.a:event)
+  let cur_buffer = gitgutter#utility#bufnr()
+
+  call gitgutter#utility#set_buffer(self.buffer)
+  call gitgutter#debug#log('job_id: '.a:job_id.', event: '.a:event.', buffer: '.self.buffer)
 
   if a:event == 'stdout'
     " a:data is a list
@@ -58,6 +62,8 @@ function! gitgutter#async#handle_diff_job_nvim(job_id, data, event)
     call s:job_finished(a:job_id)
 
   endif
+
+  call gitgutter#utility#set_buffer(cur_buffer)
 endfunction
 
 


airblade:
Thanks!  This problem crossed my mind at one point but then I forgot about it.

I'm just heading out for the weekend so I'll look at this on Tuesday (Monday's a holiday here).

tweekmonster:
No problem.  Thanks for the courtesy and enjoy the holiday!

airblade:
I wonder what happens if two jobs call back at the same time for different buffers, given that `gitgutter#utility#bufnr()` is global.  I doubt the code is reentrant.  I expect we could still get signs applied to the wrong buffer.

If that's the case, I could either pass the bufnr into each and every function that needs it so it's no longer a global (though I'm not convinced that would be thread-safe), or implement some kind of lock (which sounds like overkill).

Overall I believe your patch would improve the situation but I'd be interested to hear what you think.

tweekmonster:
I'm not that familiar with Neovim's async internals, so what follows could be wrong.

From my observation, the only part that's async is the external process when spawned.  Neovim will block once processing is back in VimL's context, including the processing of pending async jobs.  I assume that this is the case with mainstream Vim's new job control as well.  If VimL was truly asynchronous, I would expect that locking mechanisms were available to avoid race conditions.  Your concern may be a non-issue as long as the current buffer is restored at the end of script execution.

I did a small test with the following:

```diff
--- a/autoload/gitgutter/async.vim
+++ b/autoload/gitgutter/async.vim
@@ -37,12 +37,20 @@ function! gitgutter#async#execute(cmd)
   endif
 endfunction

+function! s:randnum(max) abort
+  return str2nr(matchstr(reltimestr(reltime()), '\v\.@<=\d+')[1:]) % a:max
+endfunction

 function! gitgutter#async#handle_diff_job_nvim(job_id, data, event)
   let cur_buffer = gitgutter#utility#bufnr()

   call gitgutter#utility#set_buffer(self.buffer)
   call gitgutter#debug#log('job_id: '.a:job_id.', event: '.a:event.', buffer: '.self.buffer)
+  let t = s:randnum(500)
+  let time = reltimestr(reltime())
+  echomsg 'sleep: '.t.'ms - time: '.time
+  execute 'sleep '.t.'m'
+  echomsg 'job_id: '.a:job_id.', event: '.a:event.', buffer: '.self.buffer.', time: '.time

   if a:event == 'stdout'
     " a:data is a list
```

Using a "random" sleep so that the printed timestamps would be out of order if they were being processed asynchronously.

This is the output after restoring a two buffer session in Neovim:

```
sleep: 260ms - time: 1138049.420780
job_id: 1, event: stdout, buffer: 2, time: 1138049.420780
sleep: 360ms - time: 1138049.707378
job_id: 1, event: stdout, buffer: 2, time: 1138049.707378
sleep: 311ms - time: 1138050.113340
job_id: 2, event: stdout, buffer: 3, time: 1138050.113340
sleep: 92ms - time: 1138050.475661
job_id: 3, event: stdout, buffer: 2, time: 1138050.475661
sleep: 80ms - time: 1138050.582101
job_id: 3, event: stdout, buffer: 2, time: 1138050.582101
sleep: 114ms - time: 1138050.716162
job_id: 4, event: stdout, buffer: 2, time: 1138050.716162
sleep: 221ms - time: 1138050.844741
job_id: 4, event: stdout, buffer: 2, time: 1138050.844741
sleep: 23ms - time: 1138051.099077
job_id: 7, event: stdout, buffer: 3, time: 1138051.099077
sleep: 54ms - time: 1138051.136576
job_id: 6, event: stdout, buffer: 2, time: 1138051.136576
sleep: 46ms - time: 1138051.205063
job_id: 5, event: stdout, buffer: 3, time: 1138051.205063
sleep: 286ms - time: 1138051.264307
job_id: 6, event: stdout, buffer: 2, time: 1138051.264307
sleep: 360ms - time: 1138051.585883
job_id: 1, event: exit, buffer: 2, time: 1138051.585883
sleep: 2ms - time: 1138051.946532
job_id: 2, event: exit, buffer: 3, time: 1138051.946532
sleep: 269ms - time: 1138051.948791
job_id: 3, event: exit, buffer: 2, time: 1138051.948791
sleep: 304ms - time: 1138052.218336
job_id: 4, event: exit, buffer: 2, time: 1138052.218336
sleep: 263ms - time: 1138052.522784
job_id: 7, event: exit, buffer: 3, time: 1138052.522784
sleep: 37ms - time: 1138052.786560
job_id: 5, event: exit, buffer: 3, time: 1138052.786560
sleep: 283ms - time: 1138052.823804
job_id: 6, event: exit, buffer: 2, time: 1138052.823804
```

This was the autocmd's in use to explain the number of messages above:

```vim
autocmd SessionLoadPost * GitGutterAll
autocmd BufEnter * GitGutter
```

airblade:
Thanks for researching this.

> From my observation, the only part that's async is the external process when spawned. Neovim will block once processing is back in VimL's context, including the processing of pending async jobs.

That would make things much easier.

I've noticed some buggy behaviour testing this.  I added a `sleep 2` to all commands and split the window to edit a file in each half.  I can't quite come up with solid repro steps, though I can eventually reproduce the problem every time, but if I make a change in one file and jump quickly to the other buffer, and keep doing this, eventually I get what appears to be signs for one buffer diffed against the other (or rather the version of the other in the index, for realtime diffs).

> I assume that this is the case with mainstream Vim's new job control as well. 

Who knows ;)  Vim's async stuff keeps segfaulting when vim-gitgutter uses it.

airblade:
I think the problem is the line where [the current buffer is written to a temporary file](https://github.com/airblade/vim-gitgutter/blob/f03a1a1f021ae1f4d04b6a2dc533f3329a5fe7a4/autoload/gitgutter/diff.vim#L86).

It seems that sometimes the current buffer isn't what it was earlier in the `run_diff()` function.

airblade:
I applied your patch in a4e75dd.

I'd like to figure out a solution the buggy behaviour I saw but that's a separate matter.

tweekmonster:
> I applied your patch in a4e75dd.

Thanks!

> I'd like to figure out a solution the buggy behaviour I saw but that's a separate matter.

I guess the best option would be having all relevant functions require a buffer number as an argument, as you said.  Though, I'm not sure how you would pass context variables using Vim's job functions.  Perhaps a simple mutex implemented in Vim to keep the changes minimal?  This is totally untested:

```vim
let s:lockfile = tempname()

function! mutex#acquire() abort
  let wait = 0
  while filewritable(s:lockfile)
    if wait >= 500
      " ~5 seconds
      throw '[mutex] acquire timeout'
    endif
    sleep 10m
    let wait += 1
  endwhile

  call writefile([''], s:lockfile, 'b')
endfunction

function! mutex#release() abort
  if !filewritable(s:lockfile)
    throw '[mutex] release before acquire'
  endif

  if delete(s:lockfile) == -1
    throw '[mutex] could not release'
  endif
endfunction
```

Issue #324, opened 2016-04-21
Signs disappear upon saving in neovim


tlarrieu:
Observed behavior : 
- when I save a buffer, the signs disappear and I can't navigate hunks
- when I make a modifications, signs are back (until I save the buffer)
- switching focus between splits has no effect whatsoever

This commit appears to be the culprit : 103acc7a2337c52389ad1e4b4653f5dcaaff45dc (found with git bisect)

---

My setup is as follows :

```
$ git show HEAD -s
commit 035ea9260c40bac6fee8db013522353d019dbe8d
Author: Andy Stewart <boss@airbladesoftware.com>
Date:   Thu Apr 21 15:57:07 2016 +0100

    Add section on extensions to README.
```

```
$ nvim --version
NVIM v0.1.4-179-ga7a0bf5
Build type: Dev
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -Wconversion -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -D_GNU_SOURCE -I/tmp/yaourt-tmp-tlarrieu/aur-neovim-git/src/neovim-git/build/config -I/tmp/yaourt-tmp-tlarrieu/aur-neovim-git/src/neovim-git/src -I/usr/include -I/usr/include -I/usr/include/luajit-2.0 -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/tmp/yaourt-tmp-tlarrieu/aur-neovim-git/src/neovim-git/build/src/nvim/auto -I/tmp/yaourt-tmp-tlarrieu/aur-neovim-git/src/neovim-git/build/include
Compiled by tlarrieu@tlarrieu-pc

Optional features included (+) or not (-): +acl   +iconv    +jemalloc 
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"
```

bugeats:
I've got the same. `NVIM v0.1.4-179-ga7a0bf5`.

airblade:
[...]

PR #322, opened 2016-04-18
Allow spaces and quoting in the &shell option

gwerbin:
Closes #321

Issue #321, opened 2016-04-17
Neovim: gitgutter breaks when a flag is added to `&shell`


gwerbin:
In order to get my `.zprofile` to run in the Neovim terminal, I have this line in my `nvim/init.vim` file:

```
let &shell='"/usr/local/bin/zsh" -l'
```

Gitgutter _really_ doesn't like it when I do this. If the plugin is loaded, any time I enter a normal-mode command in a file being tracked by Git I get this error:
```
Error detected while processing function gitgutter#process_buffer[8]..gitgutter#diff#run_diff:
line   75:
E902: ""/usr/local/bin/zsh" -l" is not an executable
```

The output from `:GitGutterDebug` seems okay:
```
NVIM 0.1.4-dev
Build type: Dev
Compilation: /usr/local/Library/ENV/4.3/clang -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -I/tmp/neovim20160415-59562-1w6hc5b/build/config -I/tmp/neovim20160415-59562-1w6hc5b/src -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include/luajit-2.0 -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/deps-build/usr/include -I/usr/local/opt/gettext/include -I/usr/include -I/usr/include -I/tmp/neovim20160415-59562-1w6hc5b/build/src/nvim/auto -I/tmp/neovim20160415-59562-1w6hc5b/build/include

git version 2.8.1

grep (BSD grep) 2.5.1-FreeBSD
[...]

PR #333, opened 2016-05-07
Fix Neovim jobstart() invocation

aymericbeaumet:
[Neovim's jobstart()](https://neovim.io/doc/user/eval.html#jobstart()) directly
executes the command when a list of arguments is passed. This commit makes sure
providing a space in &shell or &shellcmdflag is supported.

This needs to be explicitly supported, otherwise this generates an error (binary
not found):

```vimscript
set shell=bash\ -l
```

### DIFF
--- a/autoload/gitgutter/async.vim
+++ b/autoload/gitgutter/async.vim
@@ -9,7 +9,8 @@ endfunction
 
 function! gitgutter#async#execute(cmd)
   if has('nvim')
-    let job_id = jobstart([&shell, &shellcmdflag, a:cmd], {
+    let command = split(&shell) + split(&shellcmdflag) + [a:cmd]
+    let job_id = jobstart(command, {
           \ 'on_stdout': function('gitgutter#async#handle_diff_job_nvim'),
           \ 'on_stderr': function('gitgutter#async#handle_diff_job_nvim'),
           \ 'on_exit':   function('gitgutter#async#handle_diff_job_nvim')


airblade:
Thanks for this.  We had something similar in #321 / #322 but it broke things for a few people (#324) even though it worked for me.

@gwerbin @tlarrieu @bugeats @martin-svk @yyadavalli @adambiggs Would you mind trying out this change and let me know if it breaks anything?  Thanks.

gwerbin:
@aymericbaumet after I proposed this change, I have learned  that that is the incorrect way to invoke the shell with flags. The correct way would be to `set shell=bash` and then `set shellcmdflags=-l`. That said, in the original implementation, the call to `jobstart` omitted the second piece.

tlarrieu:
I'll give it a shot later today (when I'm at home basically :) )

aymericbeaumet:
This [PR](https://github.com/airblade/vim-gitgutter/pull/322/files) approach (to the same problem) handles quotes, while this one doesn't. I merely took inspiration from the neovim help, I think this is the way to go, just need to make sure quotes are dealt with properly.

tlarrieu:
From what I can tell, it seems to work properly.

I have not worked with this for an extensive period of time but nothing went wrong for the one-ish hour I spent with it.

bugeats:
So far so good today.

aymericbeaumet:
@airblade should we move forward?

airblade:
@aymericbeaumet I'm still not sure about this.  The problem you want to solve is `set shell=bash\ -l` causing an error.  However as @gwerbin pointed out, that's the wrong way to do it.  You should have `set shell=bash` and `set shellcmdflag="-l -c"`.

I got the list version of the `jobstart()` command from the examples in the [job control usage](https://neovim.io/doc/user/job_control.html#job-control-usage) docs.

The `jobstart()` docs say:

```
jobstart({cmd}[, {opts}])				{Nvim} *jobstart()*
		Spawns {cmd} as a job.  If {cmd} is a |List|, it will be run
		directly.  If {cmd} is a |string|, it will be roughly 
		equivalent to
			:call jobstart(split(&shell) + split(&shellcmdflag) + ['{cmd}'])
 		NOTE: read |shell-unquoting| before constructing any lists 
		with 'shell' or 'shellcmdflag' options.  The above call is 
		only written to show the idea, one needs to perform unquoting 
		and do split taking quotes into account.
```

That NOTE seems a bit of a cop-out to me:

> ...one needs to perform unquoting and do split taking quotes into account.

This sounds (a) tricky and (b) like there's a right way to do it, so shouldn't it be built into NeoVim?  If we adopt the `jobstart(split(&shell) + split(&shellcmdflag) + ['{cmd}'])` then, according to the docs, this will break in the face of quotes.

I think I'll open a NeoVim issue requesting further information.

In the meantime, if you do `set shell=bash` and `set shellcmdflag="-l -c"`, does vim-gitgutter work for you?

airblade:
There's a discussion about `jobstart()` versus `system()` and how they handle arguments in neovim/neovim#2143.

airblade:
I just asked the best way to call `jobstart()` in neovim/neovim#4746.

airblade:
Does this work for you?

```diff
-    let job_id = jobstart([&shell, &shellcmdflag, a:cmd], {
+    let job_id = jobstart(a:cmd, {
```

The discussion in neovim/neovim#4746 suggests this might be the best way.

airblade:
I believe 1937f06 should fix the problem this pull request solves...

gwerbin:
@airblade I saw the conversation on the Neovim repo, so I understand why you're hardcoding Bash in here.  But in that case the user might run into PATH issues finding Git, no? Maybe the user ought to be able to specify the path to the correct Git executable.

airblade:
@gwerbin Good idea: f5a5234.

PR #338, opened 2016-05-17
Support VIM older than 7.04

Yasumoto:
The `getbufvar` call in 7.04 now takes a 3rd argument to optionally
return a default value. This maintains that ability, but puts in logic
to safely fix this on older vim installations.


airblade:
Many thanks – however I found a more compact solution in f15a67f.

Yasumoto:
Even better- thank you!

PR #341, opened 2016-05-18
Fix typo in table of contents

Saser:
See title.


airblade:
Thanks!  It's in b3979e7.

PR #342, opened 2016-05-18
Use /bin/sh as default shell to maintain OpenBSD compatibility

xguerin:
Changed "/bin/bash" calls to "/bin/sh" as bash is not part of OpenBSD base system.  Tested both gutter and async operations.


airblade:
Thanks!  It's in 377c71d.

PR #343, opened 2016-05-18
Define g:gitgutter_shell_command

xguerin:
So that other default shells than `/bin/bash` can be used, especially on systems that do not have `/bin/bash`.


airblade:
If we go down this route we may as well just use `&shell`, leaving us at the mercy of the user's configuration again.

Closing in favour of #342.

PR #354, opened 2016-06-28
add instructions on installing with Vundle

Strangehill:


xohzzwn6kcj9:
+1

airblade:
I don't think this is really necessary.

PR #357, opened 2016-07-12
Use "ch_info()" to get channel id

fszymanski:


airblade:
Thanks for this.  I see that `ch_info()` was added to Vim in 7.4.1624 so I'm not sure how I missed it then.  It's good to have it now :)

PR #358, opened 2016-07-14
Check if job failed to start in Vim

fszymanski:


airblade:
Thanks for this.

Have you ever seen a job fail?

fszymanski:
Honestly, I did not.
In Neovim we have `job_id < 1` so I added this just in case.

airblade:
I'm happy to take a chance without this ;)

PR #383, opened 2016-10-27
Replace "normal" command with "normal!"

tlarrieu:
We do not want to use user defined mappings


airblade:
Thank you!

tlarrieu:
My pleasure :)

PR #393, opened 2016-12-13
Declare nvim's callback as a dict function

### LINKED ISSUE #392: Breaks in Neovim after partials commit
`Error detected while processing function gitgutter#async#handle_diff_job_nvim:
line    1:
E121: Undefined variable: self`.

I believe it's related to the commit of the partials/timers [patches](https://github.com/neovim/neovim/pull/5529).

jamessan:
neovim/neovim#5529 merged support for Vim's partial functions, which
made nvim more strict about dictionary functions.  This means the
implicit self variable isn't available unless a function is annoted as a
dict function or defined as part of a dictionary.

Closes #392

### DIFF
--- a/autoload/gitgutter/async.vim
+++ b/autoload/gitgutter/async.vim
@@ -79,7 +79,7 @@ function! gitgutter#async#execute(cmd) abort
 endfunction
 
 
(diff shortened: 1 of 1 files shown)


airblade:
Thank you!

Issue #328, opened 2016-04-24
Error thrown when saving to a network path on Windows


kflu:
Sorry I can't reopen #61, but I'm still seeing this with the latest master branch (5ed00d1bd386e19f5bf2996fedd4021760e99591). 

More specifically, on opening a Windows network share file (`\\share\folder\file.txt`), it prompts the error

```
Error detected while processing function GitGutter..<SNR>60_is_active..<SNR>60_is_in_a_git_repo:
line 2:
E484: Can't open file C:\Users\ljani\AppData\Local\Temp\VIoEE6D.tmp
```

And it gets periodically popped up every few seconds.

airblade:
Does it help if you change the start of [this line](https://github.com/airblade/vim-gitgutter/blob/4bed1da6486f26324f97307e0f20574af0f549ea/autoload/gitgutter/utility.vim#L152) from `'cd '` to `'cd /d '`?

kflu:
No, still given me the error. But I just realized the error message might be slightly different than what was given in #61. Here's the one I'm seeing:

![untitled](https://cloud.githubusercontent.com/assets/1031978/14837085/4cff6f04-0bca-11e6-971a-6b7f04e71caf.png)

airblade:
That error message means that Windows wasn't able to execute the external command for some reason, presumably the network share path.

I'm not familiar with Windows or network shares so I don't know how to adjust for the different path.

You could upgrade to the latest vim-gitgutter, do `:let g:gitgutter_log=1`, then reproduce the error.  If you look in the vim-gitgutter directory you'll see a `gitgutter.log`...please could you post the contents here?  That will show the exact command that is failing.

kflu:
Thanks for looking into this @airblade here you go:

```
function gitgutter#process_buffer[8]..gitgutter#diff#run_diff[90]..gitgutter#utility#system[1]:
cd "\\share\public\Files\UserFiles\kfl" && (git ls-files --error-unmatch Clean-HostsCert.ps1 && (git -c "diff.autorefreshindex=0" diff --no-ext-diff --no-color -U0   -- Clean-HostsCert.ps1))

function gitgutter#process_buffer[8]..gitgutter#diff#run_diff[13]..gitgutter#utility#file_relative_to_repo_root[3]..gitgutter#utility#system[1]:
cd "\\share\public\Files\UserFiles\kfl" && git rev-parse --show-prefix
```

My repro step was to edit a file on the share in vim with this command:

[...]

PR #400, opened 2016-12-27
#328 ignore exception when opening files on windows network shares

Snaptags:
see my post in #328.


airblade:
I'm closing this in the hope we'll find a solution in #328.

PR #401, opened 2016-12-30
Detect git client version >= 1.7.2 without using a regular expression

jssjr:
We use a custom version of git at work and our modified version string breaks git_gutter's detection of a git 1.7.2+ client. By borrowing from https://github.com/git/git/blob/35f6318d44379452d8d33e880d8df0267b4a0cd0/contrib/hooks/multimail/git_multimail.py#L416-L432 it's possible to determine if we support overriding config on the command line without relying on a regex. This should be more reliable and remove some unnecessary code.

My vimscript proficiency is quite poor, so any suggestions on writing this in a more idiomatic manner would be appreciated. :bow:

### DIFF
--- a/autoload/gitgutter/utility.vim
+++ b/autoload/gitgutter/utility.vim
@@ -177,7 +177,7 @@ endfunction
 " True for git v1.7.2+.
 function! gitgutter#utility#git_supports_command_line_config_override() abort
   call system(g:gitgutter_git_executable.' -c foo.bar=baz --version')
-  return v:shell_error == 0
+  return !v:shell_error
 endfunction
 
 function! gitgutter#utility#stringify(list) abort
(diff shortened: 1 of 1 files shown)


airblade:
Thanks!  I agree with you: feature detection is better than parsing a version string.

I also appreciate the link to the code in git's own repo.

Please could you add `call` before `system(...)`?  I.e. `call system(...)`.  You can only execute methods without `call` when they are on the right hand side.

And you can make the return value more idiomatic with `return !v:shell_error` because non-zero values are truthy in vim, and `v:shell_error` is zero for a successful command.

jssjr:
👍 Thanks for the feedback. I've updated the PR to include the changes you proposed.

airblade:
Thanks again!

PR #407, opened 2017-02-28
Don't write file when it's impossible to do so.

PhilRunninger:
I wrote a bash script that uses vim to open a text file. It uses the `-m` switch, which prevents vim from being able to write files. It also causes gitgutter to throw errors, specifically on this [line](https://github.com/airblade/vim-gitgutter/blob/master/autoload/gitgutter/diff.vim#L88).

I simply added an `if` statement around it to prevent vim from trying to write if the settings disallow it.


airblade:
Thanks!  Fixed in d137753.

PR #419, opened 2017-04-29
External commands (`:!`) update the gutter

### LINKED ISSUE #418: `:!git add %` doesn't update gitgutter
To reproduce:

1. make a change in a file, wait for gitgutter update
2. `:!git add %` to stage the current file
3. gitgutter doesn't update until I `:e`, even when `autoread` is set

tylerbrazier:
Fixes #418


airblade:
Thank you!

PR #422, opened 2017-05-05
Improve support for altering base of diffs

aschrab:
Allow per-buffer configuration of base used for diffs and add `GitGutterBase` command to set the base and update the display.

I open up files from pull requests within vim to more closely examine the changes, and in those cases I want to use the PR target as the base for diffs to help with jumping between hunks. But, when I'm writing code the default of diffing against the index (or possibly HEAD) is more useful. Having a single setting for the base across the entire vim session can make this more difficult, also having a command to change the setting makes it easier to use tab completion.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -256,4 +256,16 @@ function! gitgutter#preview_hunk() abort
   call gitgutter#utility#restore_shell()
 endfunction
 
+function! gitgutter#set_diff_base(base)
+  let b:gitgutter_diff_base = a:base
+endfunction
+
+function! gitgutter#get_diff_base()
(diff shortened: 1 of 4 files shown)


airblade:
Thanks for this.  I'm a bit snowed under at the moment but I'll try to look at this properly soon...

airblade:
I'm going to pass on this.  I can see how it would be useful, but personally I would find it a bit confusing having buffers diffing against different bases.  I'm not convinced there's a lot of demand for it either.

I'm more persuaded by having a tab-completable command, but it's still a thin wrapper.  Having the base itself tab-completable would be nice but it's probably non-trivial.

PR #424, opened 2017-06-16
Add option to disable deprecated map.

doronbehar:
I didn't like a mapping I don't use when I `nnoremap <leader>` so I made this PR,
what do you think?


airblade:
I removed the deprecation warning altogether (a8af185).  It's been there long enough.

PR #444, opened 2017-10-31
Open folds on hunk jumping

sudo-nice:
Hello!
This small feature will allow us to get folds unwrapped when we jump to hunks.


airblade:
Thanks!  It's in 1e60667.

PR #451, opened 2017-12-08
Improve hunk preview generation

somini:
- Don't generate preview windows with empty lines. Treat
`&previewheight` as a maximum value, open smaller windows for smaller
diffs. This also eliminates the extra empty line that the preview
created.
- Keep the preview buffer unmodified after inserting the diff.

This is an improvement over https://github.com/somini/vim-gitgutter/commit/e648217c28a63ddf8b08546942a0462d6bf6d395

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -239,16 +239,18 @@ function! gitgutter#preview_hunk() abort
       call gitgutter#utility#warn('cursor is not in a hunk')
     else
       let diff_for_hunk = gitgutter#diff#generate_diff_for_hunk(diff, 'preview')
+      let lines_for_hunk = split(diff_for_hunk, "\n")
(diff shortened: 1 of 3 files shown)


somini:
I added some text to the statusline because I was constantly checking the diff fit in the preview window.

Now that I think about it, this should default to false or it will break all statusline plugins.

somini:
I'm using this for a while now, and the only "bug" I found was when diffing long lines, if the `wrap` option is set, the visible lines are not enough to show the entire diff, due to some text lines using more than a visual line.
Not really a bug, just a limitation to be aware of.

somini:
Rebased this to the latest master.

airblade:
I brought in the height changes but left out the statusline stuff.

Thanks!

PR #452, opened 2017-12-09
Provide a way to see staged changes [Redux]

somini:
This is PR #257 rebased to the current repo state.

Even though much has changed, this wasn't too hard to rebase.

TODO items just like last time:
- [ ] Hunk movement in staged mode
  Since the staged hunks are calculated by cheating, they are not set to the correct places
- [ ] Make sure the tests pass
  Running the tests shows many failures, but I can't interpret them correctly.

"Staging" and "Reverting" hunks in staged mode might be too difficult to implement, let alone design a sane UI, so I have stubbed those out, they are not available in staged mode.


somini:
The tests on `master` also fail, since I'm using Neovim.

somini:
It seems the test methodology changed, that's why my test is not running.

somini:
Rebased this to the latest master.

airblade:
I appreciate all the work you've done on this, but I think this is out of scope for gitgutter.  The plugin is designed to show differences between the working tree and the index; showing differences between the index and head is not a goal.

somini:
622193a625b6f26622b9821933f9424ccf0b6fd8

Yay!

PR #471, opened 2018-02-15
Fix corrupted characters in latin1 terminals

h3xx:
This fix at least makes it so if [vim-airline](https://github.com/bling/vim-airline) and [cream-showinvisibles](https://github.com/albfan/cream-showinvisibles) don't show unicode characters, neither will vim-gitgutter.

Based on the settings of `&encoding` and `has('gui_running')` we should decide whether to show UTF-8 characters. These settings get set [and sometimes reset after your vimrc loads] automagically based on:

* how Vim interprets what you set your TERM environment variable to
* whether the value of LC_MESSAGES =~ /\.utf8/
* whether the value of LC_CTYPE =~ /\.utf8/
* evil spirits summoned through the value of TERMCAP

I don't understand it fully, but I tested it in 3 environments and this fix works in all 3 of them.

### DIFF
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -42,11 +42,11 @@ call s:set('g:gitgutter_override_sign_column_highlight', 1)
 call s:set('g:gitgutter_sign_added',                '+')
 call s:set('g:gitgutter_sign_modified',             '~')
 call s:set('g:gitgutter_sign_removed',              '_')
-try
+if has("gui_running") || &encoding == "utf-8"
   call s:set('g:gitgutter_sign_removed_first_line', '‾')
-catch /E239/
-  let g:gitgutter_sign_removed_first_line = '_^'
-endtry
+else
+  call s:set('g:gitgutter_sign_removed_first_line', '_^')
+endif
 
 call s:set('g:gitgutter_sign_modified_removed',    '~_')
 call s:set('g:gitgutter_diff_args',                  '')


airblade:
I've done some experiments with an overbar/overscore as a sign and in the buffer text.  The buffer text is irrelevant here but I included it for completeness.

### Terminal vim (both `mvim -v` and vim on ubuntu)

| `encoding` | `termencoding` | sign | buffer |
|--|--|--|--|
| utf-8 | | ✓ | ✓ |
| latin1 | | ✓ | x |
| utf-8 | latin1 | x | x |
| latin1 | latin1 | ✓ | x |
| utf-8 | utf-8 | ✓ | ✓ |
| latin1 | utf-8 | x | x |

=> ok if `termencoding == ''` or `termencoding == encoding`

### MacVim GUI (`termencoding` is hard-coded to utf-8)

| `encoding` | `termencoding` | sign | buffer |
|--|--|--|--|
| utf-8 | utf-8 | ✓ | ✓ |
| latin1 | utf-8 | x | x |

=> ok if `encoding == 'utf-8`; alternatively, if `termencoding == encoding`

### Gvim (Windows)

| `encoding` | `termencoding` | sign | buffer |
|--|--|--|--|
| utf-8 | | ✓ | ✓ |
| latin1 | | x | x |
| utf-8 | latin1 | ✓ | ✓ |
| latin1 | latin1 | x | x |
| utf-8 | utf-8 | ✓ | ✓ |
| latin1 | utf-8 | x | x |

=> ok if `encoding == 'utf-8'`

airblade:
The results give us:

```viml
if windows
  return &encoding ==# 'utf-8'
else
  return &termencoding == &encoding || &termencoding == ''
endif
```

h3xx:
Thanks for the thorough think-through and testing!

h3xx:
Unfortunately this doesn't fix my environment:

| &termencoding | &encoding | sign
| --- | --- | ---
| latin1 | latin1 | `¿`
| [*]  | latin1 | `¿`

Expected sign: `^_`.

[*] After removing the `&termencoding` [hack from my vimrc](https://github.com/h3xx/dotfiles/blob/master/vim/.vimrc#L19).

airblade:
Are you running GUI vim or terminal?  And on some kind of linux?

PR #473, opened 2018-02-17
Trigger update when the buffer is written

nathanrossi:
It appears that 5bfe5b92099e0d332a7ca26c0263fab509ee9617 removed the BufWritePost autocmd that triggers processing when the buffer is written. The commit did not mention any reason for its removal (maybe I am missing something?). However it is useful to have this behavior such that it is not required to set updatetime to a lower value only to trigger CursorHold autocmds when it can instead be triggered earlier when the user explicitly updates the file.

### DIFF
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -218,6 +218,7 @@ augroup gitgutter
         \   call gitgutter#process_buffer(bufnr(''), 0) |
         \ endif
 
+  autocmd BufWritePost                      * call gitgutter#process_buffer(bufnr(''), 0)
   autocmd CursorHold,CursorHoldI            * call gitgutter#process_buffer(bufnr(''), 0)
   autocmd FileChangedShellPost,ShellCmdPost * call gitgutter#process_buffer(bufnr(''), 1)
 


airblade:
I removed `BufWritePost` because whether or not the buffer is saved makes no difference to the diff between the buffer and the index.

Why do you not want to reduce `updatetime`?

nathanrossi:
Its mostly that I am used to writing the file triggering the updates (which I have configured the same for other plugins). I was never using the `gitgutter_realtime` option previously, so the change in behavior was a bit jarring. Whilst I don't really mind setting `updatetime` to a low value, it is more that I am less interested in the feature of showing the diff between the buffer vs index compared to the working tree vs index. But this is just my preference, and the existing code sort of enabled that behavior.

I will try out using the `updatetime` reduced as suggested, and if I don't get accustomed to the buffer diff I will sort myself out with some config in my vimrc. Either way thanks for your work, this great plugin and taking the time to review this PR.

airblade:
I imagine the change in behaviour would be a little jarring; hopefully it's a good change though.

And, as you say, you can always add that autocmd to your own vimrc.

PR #474, opened 2018-02-18
Fixed call to current_hunk in hunk text object

geneotech:
It appears that the ``s:current_hunk`` has been changed to take no argument, but the ``gitgutter#hunk#text_object`` still calls it with the result of ``bufnr('')``. This breaks some useful combos like ``vic``, which selects the entire hunk.

This commit fixes the call, restoring the intended behaviour of the hunk text object.


airblade:
Thanks!  It's in 314e646.

airblade:
Added tests in ef94b19.

PR #477, opened 2018-02-20
Improve GitGutter User autocommand: context/unsilent

blueyed:
This provides `g:gitgutter_hook_context` during the hook's execution and
removes the `:silent`, but uses `exists()` instead.

The bufnr might be necessary to know in the User autocommand, e.g. to
clear some cache.
Not using`:silent` is good practice in general to not hide (wanted)
output and errors etc.


airblade:
That's definitely an improvement – thank you!

It's in a046b5a.

PR #479, opened 2018-02-20
gitgutter#utility#setbufvar: skip unnecessary setbufvar

blueyed:
An existing dict gets returned by reference, so changing it does not
need a call to `setbufvar` afterwards.


airblade:
Good point.  Thanks for the patch.

It's in 18f6d0f.

PR #480, opened 2018-02-20
Conditionally define s:vimdiff

blueyed:
This optimizes it a bit since the `exists()` is not called always.


airblade:
Interesting – I never thought of defining functions conditionally but of course there's no reason not to.

I learned something new :)

It's in 4931dd1.

Issue #449, opened 2017-11-27
Error with Dedupe(list)


AdeelK93:
Hi,

I'm using your plugin in conjunction with the [ultimate vimrc setup](https://github.com/amix/vimrc) and I appear to be getting an error from [this part of the dedup code](https://github.com/airblade/vim-gitgutter/blob/1c034be0d31168c8f4770ef7b69adb67d00d6f3d/autoload/gitgutter/utility.vim#L187).

The error I'm getting is:
```
E121: Undefined variable: v:key
E116: Invalid functions for index(a:list, v:val, v:key + 1) == -1
```

I'm not entirely sure what list this is deduping, but removing the two calls to this function from gitgutter.vim and replacing them with the undeduped list itself doesn't seem to be causing any errors, and shows the proper git changes in the gutter.

Any idea what might be causing this? Thank you!

airblade:
The `dedup()` function is used to when the plugin is updating all the visible windows.  We need a list of the windows' buffers.  As we only want to update each buffer once, and there may be more than one window onto the buffer in a tab page, we need to dedup the list from `tabpagebuflist()`.

For example, open a file in Vim then split the window.  `:echo tabpagebuflist()` will list the buffer twice.  If you only want to process that buffer once, you need to dedup the list.

I don't know why you're getting the error though.  Do you get it with only gitgutter installed?

airblade:
Forgot to say: bypassing the deduplication won't cause any problems; it'll just mean you'll update each buffer as many times as it has visible windows, which is extra work if you have any.

airblade:
I have taken the deduplication out in 5bfe5b9.

PR #482, opened 2018-02-20
Skip duplicates from tabpagebuflist

blueyed:
There are two approaches in here - `uniq()` is not available in some older Vims.
I recommend to squash it (i.e. use the later approach).

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -3,7 +3,12 @@ let s:t_string = type('')
 " Primary functions {{{
 
 function! gitgutter#all(force) abort
-  for bufnr in uniq(sort(tabpagebuflist()))
+  let processed = {}
+  for bufnr in tabpagebuflist()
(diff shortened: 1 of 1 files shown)


airblade:
Until recently I was skipping duplicates from `tabpagebuflist()` (55b556f) but it caused a problem (#449) and I ended up dropping it because I thought the performance impact would be negligible (as `gitgutter#process_buffer()` only runs a diff if there are changes).

However there is in fact a race condition if the same buffer is processed asynchronously multiple times.

Therefore we do need to skip duplicates after all.  I merged your changes (4931dd1...d78184e) and then extracted the deduplication logic (8f6bbc9) so there's only one thing going on at a time in the loops.

Anyway, thanks for this!

airblade:
Forgot to mention `uniq()` was introduced in Vim 7.4.218.  I don't have a specific minimum Vim version in mind, but 7.4-compatibility would be nice.

PR #484, opened 2018-02-21
s:fix_file_references: handle diff.mnemonicPrefix

blueyed:
An alternative would be to run `git` with no user config always, i.e.
setting `GIT_CONFIG_NOSYSTEM` and `HOME=/dev/null` (different for
Windows then) in its environment.  Maybe there's an easier way / single
option for this..

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -229,7 +229,7 @@ endfunction
 function! s:fix_file_references(bufnr, hunk_diff)
   let filepath = gitgutter#utility#repo_path(a:bufnr, 0)
   let diff = a:hunk_diff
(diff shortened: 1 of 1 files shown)


airblade:
I didn't know about `diff.mnemonicPrefix`.  I think I will start using it :)

Another way to avoid it would be to [add `-c "diff.mnemonicPrefix=0"` to the diff command](https://github.com/airblade/vim-gitgutter/blob/c325a8b6d379db6bc8c248484d7ac7ffcf7bf571/autoload/gitgutter/diff.vim#L85-L88) but that wouldn't help people using older versions of git which don't support `-c`.

Anyway, I think this is a perfectly suitable fix.  It's in c325a8b.

PR #493, opened 2018-03-01
Fix gitgutter#utility#warn_once: use echom

blueyed:
This a) logs the message to be seen in `:mess` and b) avoids the
hit-ENTER prompt (which is likely caused by the message not being
logged and Vim deciding that you need to see it then).

So a) fixes b).


blueyed:
Also improves the error msg.

airblade:
Thanks!  In 2fe748b...85d762f.

PR #495, opened 2018-03-04
Ignore the arguments when testing the grep executable

tokoyami:
Without this the '--color=never' [example](https://github.com/airblade/vim-gitgutter/blob/master/doc/gitgutter.txt#L299) wouldn't work.


airblade:
Thanks!  It's in 9831934.

PR #499, opened 2018-03-07
Fix wrong signs when &encoding and &fileencoding differ

Bakudankun:
If file's encoding is different from Vim's &encoding, all lines which includes non-ascii characters show `~` sign.
I fixed it by converting before write the buffer.

I'm not sure if this is a good solution. JFYI.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -334,6 +334,9 @@ function! s:write_buffer(bufnr, file)
   if getbufvar(a:bufnr, '&fileformat') ==# 'dos'
     call map(bufcontents, 'v:val."\r"')
   endif
+  if getbufvar(a:bufnr, '&fileencoding') !=# &encoding
+    call map(bufcontents, 'iconv(v:val, &encoding, getbufvar(a:bufnr, "&fileencoding"))')
+  endif
   call writefile(bufcontents, a:file)
 endfunction
 


airblade:
Thanks for reporting this problem, and the patch.  I'll have to do some research to understand how it works...

elythyr:
Hi,

I had the same issue, I agreed with your solution.
The problem is that the buffer is copied in the default encoding, no matter it's own encoding.
If we take a loot at the /tmp directory where we can find both the copy of the index and the copy of the buffer :
`$ file /tmp/vjXYkaH/*
vjXYkaH/5.1.html: HTML document, ISO-8859 text, with very long lines
vjXYkaH/6.1.html: HTML document, UTF-8 Unicode text, with very long lines`

5.1.html is the file from the git index
6.1.html is the file copied from the buffer

Thanks to both of you for your work.

airblade:
@elythyr Thanks for chiming in.

I'm trying to reproduce this so I can write a test for it, but I can't seem to get a file written with the buffer's encoding.  This is what I'm doing:

```
Vim
:set enc=utf-8
:call writefile(['a•b'], 'foo')    " Do this to get the command in the history
:set enc=latin1
:call writefile(['a•b'], 'foo')    " I can't type this in latin1; I scroll back two commands

Shell
$ file foo
foo: UTF-8 Unicode text
```

I would have expected `foo` to be `latin1`.

Can you see what I'm doing wrong?

Bakudankun:
Vim internally has buffers and Strings in encoding of `&encoding` (cf. `:h 'encoding'`).
`writefile()` writes a list to a file without converting (from `&encoding`).
On the other hand, `:write` writes a buffer to a file in encoding of the buffer's `&fileencoding` after converting the buffer from `&encoding` to `&fileencoding` (cf. `:h 'fileencoding'`).

So, you get UTF-8 file by `writefile(['a•b'], 'foo')` when `&encoding` is utf-8, regardless of `&fileencoding`.

airblade:
Thanks, that's what I thought (though I see I was confused when I wrote my example above).  But why isn't `foo` ASCII text in this example:

```
Vim
:set enc=utf-8 fenc=latin1
:call writefile(['a•b'], 'foo')
:set enc=latin1 fenc=utf-8
:call writefile(['a•b'], 'foo')    " via up-arrow twice to original command; renders as below:
```

screen shot 2018-03-14 at 2 09 51 pm

```
Shell
$ file foo
foo: UTF-8 Unicode text
$ cat foo
a•b
```

Bakudankun:
I'm not sure, maybe `a•b` can't be converted to latin1, or `a•b` is too short to determine its encoding.

Here is a file encoded in cp932. Please use for testing.
[cp932.txt](https://github.com/airblade/vim-gitgutter/files/1811345/cp932.txt)

How to reproduce:
```
$ git init
$ git add cp932.txt
$ vim -c "e ++enc=cp932" cp932.txt    # You can find '~' at 4-7th line
```

airblade:
Wonderful, thank you!

Bakudankun:
I think the garbled text is actually `a•b` in UTF-8.
As `:h 'encoding'` says, it is not recommended to change encoding while vim is open.

airblade:
Merged in 3b2461b.

Many thanks again for your help.

PR #503, opened 2018-03-26
Add BOM support.

nagaohiroki:
<feff> is invisible from github...

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -341,6 +341,10 @@ function! s:write_buffer(bufnr, file)
     call map(bufcontents, 'iconv(v:val, &encoding, "'.fenc.'")')
   endif
 
+  if getbufvar(a:bufnr, '&bomb')
+    let bufcontents[0]='﻿'.bufcontents[0]
+  endif
+
   call writefile(bufcontents, a:file)
 endfunction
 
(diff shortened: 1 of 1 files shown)


airblade:
Thanks!

The docs for `'bomb'` say:

> 	When writing a file and the following conditions are met, a BOM (Byte
>	Order Mark) is prepended to the file:
>	- this option is on
>	- the 'binary' option is off
>	- 'fileencoding' is "utf-8", "ucs-2", "ucs-4" or one of the little/big
>	  endian variants.

Your patch tests the first condition; shouldn't we also test the other two conditions?

airblade:
Writing the buffer used to be implemented with `:write` but I changed it to use `writefile()` in 4e911819be.  `:write` was ok but had several side-effects which had to be dealt with, which was annoying.  I don't plan to revert to `:write`.

airblade:
I think all we need do is add a check to see whether `fileencoding` is one of the ones listed in the `bomb` help doc.

airblade:
Hmm, actually I think your original patch is fine :)  If `bomb` is set, then we should write it back out.

airblade:
I merged your original commit (3724e1c).  Thanks again!

nagaohiroki:
thank you very much!
my english is slow(sorry)

airblade:
Your English is better than my Japanese :)

PR #504, opened 2018-03-31
readfile/writefile binary mode

nagaohiroki:
Hi.
getbufline->writefile lost encode(bom, line).
use binary mode.


airblade:
I don't think your solution works when the buffer has unsaved changes.

nagaohiroki:
It was insufficient.. close this request.

PR #510, opened 2018-04-14
diff: fix handling of empty file

blueyed:
An empty, unchanged file was considered to be changed (a new line
added).
This patch fixes s:write_buffer to use binary mode with `writefile` to
not append a newline always, and does so manually for non-empty buffers,
according to &endofline, &binary and &fixendofline.

This is taken out of Neomake, and tested there in
https://github.com/neomake/neomake/blob/091d148b/tests/utils.vader#L713-L759.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -332,6 +332,15 @@ endfunction
 function! s:write_buffer(bufnr, file)
   let bufcontents = getbufline(a:bufnr, 1, '$')
 
+  " Special case: empty buffer; do not write an empty line in this case.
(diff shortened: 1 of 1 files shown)


airblade:
Applied in 9a23a2f.

I noticed that a buffer consisting only of a newline had a similar problem so I fixed that in 5c636b1.

airblade:
Thanks for your patch :)

airblade:
@blueyed I think binary mode turns off `expandtab` (see `:h edit-binary`) which can cause lines with tabs to show up as modified.  Have you encountered this?

PR #511, opened 2018-04-19
gitgutter#utility#setbufvar: ensure bufnr is a number with setbufvar

blueyed:
Otherwise you might get "E93: More than one match for X" if there is a
buffer with "X" in its name besides buffer number X.


airblade:
Thanks.

PR #512, opened 2018-04-19
BufEnter: do not call process_buffer if not enabled

blueyed:
Make `gitgutter#init_buffer` return 0/1 and use it in the BufEnter
script.


airblade:
On the upside, this avoids a function call.  On the downside, it introduces a conditional into the calling code, making it a tiny bit more complex.

In practical terms I don't think it makes any difference.

Was there a particular reason you suggested this?

blueyed:
Noticed it when debugging, and thought it would be good since autocommands should be optimized in general.
Let me know if I should rebase / fix the conflict.

Issue #526, opened 2018-05-29
gitgutter marks are not updated on file save


marbu:
I was using bit older commit and when I updated to current master c4301f9a103ac1429d05131e7debdac5eb6c4c23, I noticed that the gitgutter signs are not shown when I save the file (eg. via typing `:w`). The signs will appear eventually (roughly after the default 4 seconds, or almost immediatelly when tweaking `updatetime`), but when I disable realtime updates, I still see no signs when saving a file.

Is this expected behavior?

> What is the latest commit SHA in your installed vim-gitgutter?

5bfe5b92099e0d332a7ca26c0263fab509ee9617 (this is the 1st commit which causes the problem, found using git bissect)

> What vim/nvim version are you on?

vim-enhanced-8.1.020-1.fc28.x86_64

> If no signs are showing at all, what does `:echo b:gitgutter.path` give?

`README.rst` (which is a name of currently opened file)

> If no signs are showing at all, and the `path` value is a path and not `-2`, does it work with `let g:gitgutter_grep=''`?

This has no effect.

airblade:
It is expected behaviour: whether or not the buffer has unsaved changes makes no difference to the diff.

The idea is you set `updatetime` to a sensible value, e.g. `200`, and then the signs are always up to date.

airblade:
If you really want to update the signs when you save a file, add this to your vimrc:

```viml
autocmd BufWritePost * GitGutter
```

marbu:
That works, thank you for the help. I guess that I can close it now.

marbu:
Ah, I just noticed that about a week ago, someone asked the same question: https://github.com/airblade/vim-gitgutter/issues/523 Sorry for missing that. Would it make sense to update the FAQ (something like https://github.com/airblade/vim-gitgutter/pull/527)?

Issue #523, opened 2018-05-22
Add ability to refresh on save file


dylan-chong:
> What is the latest commit SHA in your installed vim-gitgutter?

a7206c41fc174f576b3f7085d48319dbdad760a5

> What vim/nvim version are you on?

NVIM v0.2.2

airblade:
Add this to your vimrc:

```viml
autocmd BufWritePost * GitGutter
```

I'm curious about why you need this.  If your `&updatetime` is a reasonable value and you haven't disabled gitgutter, the signs should always be up to date regardless of whether the buffer has unmodified changes or not.

dylan-chong:
Just in case there is any performance problem if changing updatetime. I'm gladly surpriseed that there doesn't seem to be one...

PR #527, opened 2018-05-30
Update FAQ in README: refresh on file save

marbu:
Proposal of FAQ update to include question about enabling refresh on save file when one disables automatic updates.

This question was asked in https://github.com/airblade/vim-gitgutter/issues/523 or https://github.com/airblade/vim-gitgutter/issues/526

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -470,6 +470,13 @@ Add this to your vim configuration (in an `/after/plugin` directory):
 autocmd! gitgutter CursorHold,CursorHoldI
 ```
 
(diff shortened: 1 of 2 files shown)


airblade:
I think it's a good idea to add this to the FAQ.

Please could you also update the FAQ in the [vim doc](https://github.com/airblade/vim-gitgutter/blob/master/doc/gitgutter.txt)?

marbu:
FAQ in vim doc updated.

airblade:
Thanks!

Issue #509, opened 2018-04-13
Many error messages after inserting php FQCN


greg0ire:
> What is the latest commit SHA in your installed vim-gitgutter?

5481318

> What vim/nvim version are you on?

<details>
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar  6 2018 08:24:16)
Included patches: 1-1573
Modified by <bugzilla@redhat.com>
Compiled by <bugzilla@redhat.com>
Huge version with GTK3 GUI.  Features included (+) or not (-):
+acl               +clientserver      +dialog_con_gui    +find_in_path      +keymap            +modify_fname      +multi_byte        +profile           +statusline        +textobjects       +wildignore
+arabic            +clipboard         +diff              +float             +lambda            +mouse             +multi_lang        +python/dyn        -sun_workshop      +timers            +wildmenu
+autocmd           +cmdline_compl     +digraphs          +folding           +langmap           +mouseshape        -mzscheme          +python3/dyn       +syntax            +title             +windows
-autoservername    +cmdline_hist      +dnd               -footer            +libcall           +mouse_dec         +netbeans_intg     +quickfix          +tag_binary        +toolbar           +writebackup
+balloon_eval      +cmdline_info      -ebcdic            +fork()            +linebreak         +mouse_gpm         +num64             +reltime           +tag_old_static    +user_commands     +X11
+balloon_eval_term +comments          +emacs_tags        +gettext           +lispindent        -mouse_jsbterm     +packages          +rightleft         -tag_any_white     +vertsplit         -xfontset
+browse            +conceal           +eval              -hangul_input      +listcmds          +mouse_netterm     +path_extra        +ruby/dyn          -tcl               +virtualedit       +xim
++builtin_terms    +cryptv            +ex_extra          +iconv             +localmap          +mouse_sgr         +perl/dyn          +scrollbind        +termguicolors     +visual            +xpm
+byte_offset       +cscope            +extra_search      +insert_expand     +lua/dyn           -mouse_sysmouse    +persistent_undo   +signs             +terminal          +visualextra       +xsmp_interact
+channel           +cursorbind        +farsi             +job               +menu              +mouse_urxvt       +postscript        +smartindent       +terminfo          +viminfo           +xterm_clipboard
+cindent           +cursorshape       +file_in_path      +jumplist          +mksession         +mouse_xterm       +printer           +startuptime       +termresponse      +vreplace          -xterm_save
[...]

PR #532, opened 2018-06-27
Fix gitgutter#utility#setbufvar for nonexistent buffers

blueyed:
Use a default with `getbufvar` to not get an empty string.

Fixes:

> Error detected while processing function <SNR>163_on_stderr_nvim[1]..<lambda>10[1]..gitgutter#utility#setbufvar:
> line    1:
> E712: Argument of get() must be a List or Dictionary

I can trigger it using https://github.com/AndrewRadev/linediff.vim when
closing the windows / the tab.
Likely triggered with/through Neomake.

### DIFF
--- a/autoload/gitgutter/utility.vim
+++ b/autoload/gitgutter/utility.vim
@@ -7,11 +7,12 @@ function! gitgutter#utility#supports_overscore_sign()
 endfunction
 
 function! gitgutter#utility#setbufvar(buffer, varname, val)
-  let dict = get(getbufvar(a:buffer, ''), 'gitgutter', {})
+  let buffer = +a:buffer
+  let dict = get(getbufvar(buffer, '', {}), 'gitgutter', {})
   let needs_setting = empty(dict)
   let dict[a:varname] = a:val
   if needs_setting
-    call setbufvar(+a:buffer, 'gitgutter', dict)
+    call setbufvar(buffer, 'gitgutter', dict)
   endif
 endfunction
 


airblade:
I can't reproduce this with linediff.vim, but never mind.  This sounds like #509.

I feel this change fixes the symptom rather than the underlying cause.  I would like to figure out why gitgutter is trying to do anything on a nonexistent buffer.  Do you know by any chance?

blueyed:
Yes, same issue.
According to https://github.com/airblade/vim-gitgutter/issues/509#issuecomment-382712676 it is due to being used async, which makes sense in general, since the buffer might be deleted/removed inbetween.

I could check which buffer it is really, but likely the temporary buffers that linediff uses (it happens when I used `Q`, mapped to my quit-window function:

```vim
function MyQuitWindow()                                                        
  let t = tabpagenr()
  let nb_tabs = tabpagenr('$')
  let was_last_tab = t == nb_tabs
                                 
  let prev_winid = get(w:, 'my_prev_winid', 0)
                                              
  if &ft != 'qf' && getcmdwintype() == ''
    lclose                               
  endif   
  " Turn off diff mode for all other windows.
  if &diff && !exists('w:fugitive_diff_restore')
    WindoNodelay diffoff                        
  endif                 
       
  if &ft ==# 'fzf'
    if get(get(g:, 'fzf_layout', {}), 'window', '') ==# 'enew'
      " TODO: trigger exe "normal! i\<c-c>" somehow?!         
      startinsert                                    
    endif        
  elseif &bt == 'terminal'
    " 'confirm' does not work: https://github.com/neovim/neovim/issues/4651
    " Might cause (through autocommand?!).                                 
    " > Error detected while processing function MyQuitWindow[19]..44:
    " line   17:                                                      
    " E516: No buffers were deleted: bd! 9%
    q                                      
  else
    confirm q
  endif      
       
  if ! was_last_tab && nb_tabs != tabpagenr('$') && tabpagenr() > 1
    tabprevious                                                    
  endif        
       
  if prev_winid
    call win_gotoid(prev_winid)
  endif                        
endfunction
```

airblade:
I think the underlying problem should be fixed by 64663ab.

blueyed:
Unfortunately not.

blueyed:
But it does not happen when using `:q` with linediff, but only when using my custom `Q` mapping.

blueyed:
The `windo diffoff` part from there triggers it.
Using `noautocmd windo diffoff` (or rather `noauotcmd WindoNodelay diffoff` (which is a wrapper command)) works around it.
So some autocommands are involved here.

kristijanhusak:
This is also happening for me when i use https://github.com/justinmk/vim-dirvish, after executing the command with `Z!`, which executes sh generated in tmp directory.

blueyed:
@kristijanhusak 
Does this PR fixes it for you?

kristijanhusak:
@blueyed yeah it does.

kristijanhusak:
@airblade can we get this merged? It's minor change.

airblade:
Sorry for the delay in merging.

> In general `gitgutter#utility#is_active` should consider buffers with path=-2 to be inactive though maybe already?

Currently the plugin checks for path=-2 at the start of `gitgutter#diff#run_diff`:

https://github.com/airblade/vim-gitgutter/blob/af1e674155ac4f04db42c7d07ff5c85fd4bb6129/autoload/gitgutter/diff.vim#L65-L67

This approach allows the user to `git add` a file that's already open in vim, and gitgutter will start showing signs for it.  If we move the check to `gitgutter#utility#is_active`, the user would have to wipeout the buffer and reopen the file to get signs to start showing.

I think the current approach is more user-friendly even though it makes the plugin do more work than it needs to, most of the time.

PR #534, opened 2018-06-28
Update all buffers (visible and hidden) on FocusGained events

### LINKED ISSUE #533: FocusGained event should clear each buffers' 'changedtick' variable in the cache
First, just wanted to thank you for a great plugin. I rely on it everyday and it's awesome!

There is one scenario that I think could be improved, however. When `FocusGained` triggers, it ends up calling `gitgutter#all(1)`:

https://github.com/airblade/vim-gitgutter/blob/a986ab054788776dca269d6c289b470255d54e8c/autoload/gitgutter.vim#L5-L13

Unfortunately, `tabpagebuflist()` only returns the buffers that are currently associated in visible windows on the current tab. It doesn't return *every* buffer. This is necessary when you have many files that have changes, and you git commit them in a different window. When coming back to vim and switching to a buffer that wasn't visible (that had signs last time it was open), the signs are not updated.

Instead of actually doing the diff for every buffer, is it possible to augment the above function to also clear the 'changedtick' bufvar for hidden buffers when called? This would allow `gitgutter#process_buffer()` to update the signs the next time the buffer is loaded after a `FocusGained` event (since `s:has_fresh_changes()` would return true).

johnmiked15:
Resolves #533 

The problem is that `tabpagebuflist()` will only return *visible* buffers, so `FocusGained` events do nothing to update the hidden buffers that aren't actually visible in any window. There are two ways to solve this. The quick and dirty way would be to change `gitgutter#all()` to loop through every listed buffer instead of only the visible ones in the current tab. 

The second method (which I've implemented here) is to call `s:clear()` on each buffer that isn't visible when force is true. This will cause the next `BufEnter` event to recalculate the diff after a `FocusGained` event has fired.

Both methods remove the need to separately track the tabs that need to be marked dirty, since all buffers regardless of tab will be updated.


johnmiked15:
If you prefer, I can add back the code that doesn't force update other tabs' visible buffers until that tab is entered again. I figured that my change dramatically simplifies the code, and I wouldn't expect there to be tons of visible buffers on non-active tabs for most users.

airblade:
Thanks – this is great.

There's one problem with deleting the code triggered by `TabEnter`: not everybody's terminal generates focus events, so we can't rely only on `FocusGained`.

So I think we should keep the tab code, and modify your `gitgutter#all(force)` so that it does nothing to buffers which are visible in other (i.e. not the current) tabs.

Would you agree or have I overlooked something again :) ?

johnmiked15:
Sure, that makes sense. I’ll add the tab code back and push it here in an hour or so.

johnmiked15:
Should be good now.

johnmiked15:
Actually, give me just a few minutes, I thought of some minor cleanup and optimization that can be added. I don’t think we need the tab force stuff anymore, since `s:reset_tick()` takes care of it when `FocusGained` events are working.

johnmiked15:
Update: the `s:flag_inactive_tabs()` code is no longer necessary.

When `FocusGained` events are working, `gitgutter#all(1)` will rerun the diff for visible buffers on the current tab and will reset the tick on all other tabs (whether or not they are visible on other tabs). The next time a new tab is entered, the `s:on_bufenter()` function will call `gitgutter#all(0)` which will only update buffers that haven't been updated since the last `FocusGained` event.

When `FocusGained` events are not firing, then `gitgutter#all(1)` is called each time a new tab is entered, which is what it was doing before.

airblade:
The updated logic looks good to me.  Thanks!

I think we need the `uniq` function though.  Without it, when a buffer is listed more than once the sign-updating callback processes conflict with each other.

I noticed you used `buflisted()` instead of `bufexists()` – is there a particular reason?  I've never really thought about it before.

johnmiked15:
We don't need `s:uniq()` anymore, actually. Since we are looping through the buffer numbers themselves, and only using the result of `tabpagebuflist()` as a way to verify a buffer is visible on a tab, we won't attempt to run the diff on the same buffer twice. It was necessary before because you were looping through the results of `tabpagebuflist()`, which may contain duplicates if a buffer is visible in multiple windows of a tab.

As for `buflisted`, I used it to cut down on the number of buffers that are loaded or reset:

>                             'buflisted' 'bl' 'nobuflisted' 'nobl' E85
>      'buflisted' 'bl'        boolean (default: on)
>                              local to buffer
>              When this option is set, the buffer shows up in the buffer list.  If
>              it is reset it is not used for ":bnext", "ls", the Buffers menu, etc.
>              This option is reset by Vim for buffers that are only used to remember
>              a file name or marks.  Vim sets it when starting to edit a buffer.
>              But not when moving to a buffer with ":buffer".

Buffers that are not listed do not contain actual editable files. They are internal buffers used by vim to store other metadata.

PR #552, opened 2018-08-15
Trigger refresh on VimResume event

### LINKED ISSUE #550: Include VimResume (neovim) alongside FocusGained autocommand
`gitgutter` has an autocommand listening on `FocusGained` events to trigger refresh. But this event doesn't exist for suspend/resume of the process. `neovim 0.3.0` introduced `VimResume` and `VimSuspend` events in this [PR](https://github.com/neovim/neovim/pull/8280). My workaround before this feature was to resume the process and immediately switch back and forth between the terminal and a different window, to force trigger of the `FocusGained` event.

Can we append this to the existing `FocusGained` autocommand? I believe this is neovim only and am not sure vim has any related plans for similar feature.

The workaround I've been using is to put this in my vimrc:
```
  if has('nvim-0.3.0')
      augroup gitgutter_resume
          autocmd!
          autocmd VimResume * call gitgutter#all(1)
      augroup end
  endif
```

ipwnponies:
VimResume is neovim event that is triggered when process is resumed to
foreground.

Fixes #550 

I tested this by install `ipwnponies/vim-gitgutter` via VimPlug (after uninstalling and cleaning up workaround).
```
:autocmd VimResume
--- Auto-Commands ---
gitgutter  VimResume
    *         call gitgutter#all(1)
```
To test original issue:
1. I modified a file and the gitgutter column showed unstaged changes.
1. Ctrl-z to background nvim.
1. `git add -A`
2. Resume process and gitgutter column no longer showed unstaged changes.

### DIFF
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -203,6 +203,10 @@ augroup gitgutter
 
   autocmd FocusGained,ShellCmdPost * call gitgutter#all(1)
 
+  if exists('##VimResume')
+    autocmd VimResume * call gitgutter#all(1)
+  endif
+
   autocmd ColorScheme * call gitgutter#highlight#define_sign_column_highlight() | call gitgutter#highlight#define_highlights()
 
   " Disable during :vimgrep


airblade:
Thanks!

PR #559, opened 2018-11-01
Add warning global variable

ColinKennedy:
I have a tool that is based on the [cycle through hunks](https://github.com/airblade/vim-gitgutter#cycle-through-hunks-in-all-buffers) code that is posted on vim-gitgutter's readme. It works well but each time a buffer runs out of hunks, vim-gitgutter displays a `No more hunks`warning message. I added a variable so that I could disable those warnings since I knew I didn't want to see them.

I figured I'd open a PR to possibly add it as an option if you're up for it.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -212,6 +212,7 @@ You can customise:
 * The signs' colours and symbols
 * Line highlights
 * The base of the diff
+* Extra arguments for `git` when running `git diff`
 * Extra arguments for `git diff`
 * Key mappings
 * Whether or not vim-gitgutter is on initially (defaults to on)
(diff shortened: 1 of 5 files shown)


airblade:
Thanks!

I'm happy to add this.  Some people like the warning, some don't, and vim is inconsistent with similar movement – see the discussion in #236.

I'll pull this in as soon.

ColinKennedy:
I see, thanks for the link. Normally I like the warning as well but there are a few select cases where it can get a bit spammy.

Also I just wanted to take a moment to thank you for making this awesome plug-in. I use vim-gitgutter as a back-end for a plug-in which emulates `git add -p` and it saves me time every day. If you ever take a trip across the pond to the US, I owe you a beer!

airblade:
Hi!

I just looked at this pull request again.  Rather than suppress all warnings, I'd rather have an option to turn off just the hunk navigation warnings, which are the controversial ones.

So I'll close this one and hope you'll have time to make a new one for that :)

PR #565, opened 2018-11-30
Add g:gitgutter_sign_allow_clobber to control clobbering

da-x:
I have a patched Neovim that allows having multiple signs, so that clobbering naturally does not happen, thus vim-gitgutter does not need to avoid it.

https://github.com/neovim/neovim/issues/990


airblade:
Thanks for this, much appreciated.  It's in 9bf988b.

Issue #502, opened 2018-03-23
Need to call :GitGutter after committing using fugitive


yevhen-m:
> What is the latest commit SHA in your installed vim-gitgutter?
`37bdd03d17`
> What vim/nvim version are you on?
`NVIM v0.2.2`

When I commit using fugitive, signs in gutter aren't updated. Need either to trigger `:GitGutter` or trigger "focus gained" event switching tmux panes. 
I guess I remember that before the plugin refactoring signs were updated after commit automatically.

airblade:
Before the refactor the plugin listened to `BufWritePost`, which Fugitive was emitting at my request.

Now the plugin doesn't listen to that event.  I have asked [whether Fugitive can emit `ShellCmdPost` or a custom autocommand](https://github.com/tpope/vim-fugitive/issues/503#issuecomment-374968526) instead.

yevhen-m:
@airblade thanks. Let's wait.

airblade:
As a workaround you could do this:

```viml
autocmd gitgutter BufWritePost * GitGutter
```

cc @amadeus

airblade:
@lisinge Sorry, I don't know.  Does gitgutter work otherwise for you in neovim?

lisinge:
@airblade yup! Works perfectly other than not updating the signs once fugitive is done

thisgeek:
I've had a nice experience so far with a slight tweak to @airblade's workaround:

```vim
autocmd BufWinLeave * GitGutterAll
```

Main benefit: The gutter resets as soon as `COMMIT_EDITMSG` is closed.
Disclaimer: I have not thought about whether this is a sledgehammer solution. It works for me.

micke:
@thisgeek thanks dude! Will try this out right away!

mk12:
I'm using:

```vim
autocmd BufWritePost,WinEnter * GitGutter
```

The `BufWritePost` I like to keep regardless because I want an update after save but don't want to reduce my `updatetime` from 4s. However, it would be nice if the present issue was fixed for Neovim too so that I wouldn't need the `WinEnter`.

blueyed:
It would be helpful if there were steps to reproduce this..

[...]

PR #591, opened 2019-03-06
[RFC] process_buffer: call init_buffer with a:force

blueyed:
This is required to re-check if a buffer is tracked now, e.g. after
vim-fugitive's `:Gwrite`.

Test case:

1. have an untracked file, open it
2. gitgutter will not process it, setting "path" to -2
3. call `:Gwrite` to add the file to the index
4. change the buffer
5. signs are not being placed
6. call `:Gitgutter` manually
7. although it calls `gitgutter#process_buffer` with `a:force = 1`, it
   will still not display signs, since `gitgutter#diff#run_diff` throws
   "gitgutter not tracked".

A workaround is to call `gitgutter#init_buffer` yourself manually.

Additionally/instead vim-gitgutter could also call `init_buffer` on
`BufWritePost`, which gets triggered by vim-fugitive, but also on every
other write of course.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -38,6 +38,10 @@ function! gitgutter#process_buffer(bufnr, force) abort
 
   if gitgutter#utility#is_active(a:bufnr)
     if a:force || s:has_fresh_changes(a:bufnr)
+      if a:force
+        let p = gitgutter#utility#repo_path(a:bufnr, 0)
+        call gitgutter#init_buffer(a:bufnr)
+      endif
 
       let diff = ''
       try


airblade:
The plugin does show signs when a file is added to the index outside vim, i.e. when step 3 above is: go to a terminal, add the file to the index, and go back to vim.  This is because of:

```viml
  autocmd FocusGained,ShellCmdPost * call gitgutter#all(1)
```

And `gitgutter#all(1)` calls `gitgutter#init_buffer()` along the way.

I want vim-gitgutter to update after `:Gwrite` (see #502) but ideally not by listening to `BufWritePost` (because whether or not the buffer is saved is irrelevant to gitgutter).

Anyway, I'll look at your patch.

autoload/gitgutter.vim
```diff
@@ -38,6 +38,10 @@ function! gitgutter#process_buffer(bufnr, force) abort
 
   if gitgutter#utility#is_active(a:bufnr)
     if a:force || s:has_fresh_changes(a:bufnr)
+      if a:force
+        let p = gitgutter#utility#repo_path(a:bufnr, 0)
```
airblade: Why is this line necessary?

blueyed: See the test case above..

I might have mixed up something here though with another PR (e.g. the return value), not sure currently.

airblade:
I think this has been made redundant by #593.

PR #593, opened 2019-03-06
Refactor `set_repo_path` handling in async mode

blueyed:
Please see the commits - I've tried to split it into smaller refactoring steps.

The main benefit is, that the sleep-loop gets removed, and the async's job callback calls `process_buffer` instead.

Only slightly tested.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -10,8 +10,7 @@ function! gitgutter#all(force) abort
       let file = expand('#'.bufnr.':p')
       if !empty(file)
         if index(visible, bufnr) != -1
-          call gitgutter#init_buffer(bufnr)
-          call gitgutter#process_buffer(bufnr, a:force)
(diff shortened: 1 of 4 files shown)


autoload/gitgutter.vim
```diff
@@ -22,12 +21,20 @@ endfunction
 
 
 " Finds the file's path relative to the repo root.
-function! gitgutter#init_buffer(bufnr)
+function! gitgutter#init_buffer(bufnr, ...) abort
```
blueyed: I've tried to keep the API compatible - but if that is not necessary `force` could be required as arg here.

airblade:
Thanks very much for this and sorry to have sat on it for so long.

I incorporated your refactor of the `set_repo_path()` async handler, and then used your code as inspiration for moving `init_buffer()` into `process_buffer()`.  This enables `run_diff()` not to have to poll the repo path and eliminates that problematic `sleep`.

See 5bb4f14...064a3d6.

Thanks!

PR #594, opened 2019-03-06
Only handle FocusGained after FocusLost

blueyed:


blueyed:
But I cannot really see why this is done/handled in the first place - is this for when buffers might have been changed from another terminal?

I think the normal `checktime` mechanism should handle this then already, no?!

Personally I find this overkill, and would therefore suggest to have a setting for this, in case it should be kept.

airblade:
> But I cannot really see why this is done/handled in the first place - is this for when buffers might have been changed from another terminal?

Partly that, but mainly to update the signs if files have been staged in another terminal.

Is the reason for the pull request to avoid unnecessary work when neovim starts up and triggers `FocusGained`?

blueyed:
> Is the reason for the pull request to avoid unnecessary work when neovim starts up and triggers `FocusGained`?

Yes.

PR #597, opened 2019-03-15
Updating instructions for overriding diff symbol colors

yan12125:
I think `ctermbg` and `guibg` are no longer necessary after commit fd834e48eed21cc3c3ab66779a2296a16f41cbca.

Tested on neovim 0.3.4 with wombat256mod color scheme [1].

[1] https://github.com/michalbachowski/vim-wombat256mod

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -288,12 +288,10 @@ You can either set these with `highlight GitGutterAdd {key}={arg}...` or link th
 To get vim-gitgutter's original colours (based on git-diff's colours in my terminal):
 
 ```viml
-highlight GitGutterAdd    guifg=#009900 guibg=<X> ctermfg=2 ctermbg=<Y>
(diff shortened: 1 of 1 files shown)


yan12125:
I found that my patch does not work with neovim 0.3.8 + wombat256mod + iTerm2 3.2.9 on macOS 10.14. I will check it out and reopen this pull request when I find a working version for all system configurations I can test with.

yan12125:
Seems the issue is from MacPorts - which I install neovim from. This pull request works on macOS after https://github.com/macports/macports-ports/pull/5033.

airblade:
Surely if you configure your own sign colours, e.g. `GitGutterAdd`, you still need to set the background colour to match the sign column's?

yan12125:
I checked my `.config/nvim/init.vim` again, and noticed that I have the following line:
```
highlight LineNr ctermbg=None
```
So it happens that ctermbg values are consistent everywhere. Sorry for the noises!

airblade:
No problem!

PR #609, opened 2019-05-02
Fix repeat support for GitGutterNextHunk / GitGutterPrevHunk

StarryLeo:


airblade:
I believe the repeat command `.` is for repeating changes not motions.

StarryLeo:
Ok, forget my stupid understanding. How about do the motions if no do the changes before, Is it possible?

airblade:
What changes before?

StarryLeo:
I mean have not done stage the hunk or undo for current buffer, just want to move between hunks.

airblade:
If you want to keep moving between hunks, do `]c ]c ]c`, or `3]c`.

You can use `.` to repeat the last hunk operation, e.g. staging or undoing.

I often stage a hunk, then move through the other ones in the buffer and stage each one.  The keystrokes are: `gg ]c ghs ]c . ]c .` etc.  (I use `ghs` for staging.)

StarryLeo:
I just want do:
`gg ]c . . . . . . ghs ]c . `
mean: 
`gg ]c ]c ]c ]c ]c ]c ghs ]c ghs`

airblade:
Thanks for the example – I understand what you want.

However vim-repeat only repeats changes

StarryLeo:
This is a wrong demonstration.

PR #611, opened 2019-05-12
Add g:gitgutter_preview_win_location option to control preview popup location

Jasha10:
The default preview window location is `bo`. This commit adds a global variable `g:gitgutter_preview_win_location` that enables other settings such as `to`, `bel` and `abo`.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -255,7 +255,7 @@ function! s:preview(hunk_diff)
 
   silent! wincmd P
   if !&previewwindow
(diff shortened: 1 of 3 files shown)


airblade:
Thanks for this.

I wonder whether it would be even better to remove the location specifier and just let Vim figure it out from the user's `splitbelow` and `splitright`.  Would that suit you?

Jasha10:
That would suit me fine.

airblade:
Thanks again for this (and sorry for the delay).  It's merged in 8985017.

PR #620, opened 2019-06-01
Simplify getting highlight fg/bg

blueyed:
Only tested quickly, but I am using this mechanism with Neomake already (where I am additionally handling "reverse" though - i.e. if "reverse" is used, it should use the "bg" for "fg", but that is out of scope of this PR).


airblade:
Thanks!  That looks simpler and perhaps / hopefully more robust.

I rebased it onto master and it's in 9f21e3d.

PR #621, opened 2019-06-03
gitgutter#utility#getbufvar: handle not existing buffer

blueyed:
`getbufvar(a:buffer, '')` returns '' then, resulting in an error.


blueyed:
This can also happen during `BufFilePost` processing, where the buffer appears to exist only after.

blueyed:
Not much tested, and have the feeling I submitted this before..  ah - for setbufvar: af1e674

Let me know if I should update this according to the mechanism used there etc.

airblade:
Thanks again.  This has never caused me a problem but other plugins can do all sorts of things with creating and destroying buffers.

It's in a4e8d3f.

Issue #551, opened 2018-08-14
Doesn't detect file name change after :saveas


pmcmorris:
> What is the latest commit SHA in your installed vim-gitgutter?
a986ab054788776dca269d6c289b470255d54e8c

> What vim/nvim version are you on?
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Nov 24 2016 16:44:48)
Included patches: 1-1689
Extra patches: 8.0.0056

> If no signs are showing at all, what does `:echo b:gitgutter.path` give?
Signs are showing but they are wrong.

> If no signs are showing at all, and the `path` value is a path and not `-2`, does it work with `let g:gitgutter_grep=''`?
Signs are showing but they are wrong.

After using the :saveas command, gitgutter continues to compare against the previous file instead of using the new name (which is probably a new name and not in the repo).  I can see all the +/-/~ get updated as I edit for a file not in the repo.

airblade:
This is because gitgutter caches the file's path after the first run, so when you rename the file with `:saveas` the plugin continues to use the original path in the git commands.

Maybe we could detect a `:saveas` by comparing file names in `BufWritePre` and `BufWritePost` events and using the new file path if they differ.

airblade:
Closing because this has gone quiet.  Happy to reopen if it's still a problem.

pylipp:
I'm facing the same issue and would like to help. Could you point me to where I could start listening for the `BufWrite*` events?

airblade:
You could add event handler in `plugin/gitgutter.vim`.  If the filenames differ pre and post, clear the path in `b:gitgutter.path` and force GitGutter to run.

airblade:
Thanks for letting me know.  I'll reopen this so I don't forget.

blueyed:
@pylipp 
Can you create a PR for it, please?

Or do you plan to cherry-pick / apply it manually?

For reference: relevant autocommands being emitted:

```
BufNew: abuf: 12, amatch: …/vim/plugged/vim-gitgutter/meh.vim, winnr():1, ft: vim, bt: )
BufFilePre: abuf: 1, amatch: …/vim/plugged/vim-gitgutter/bar.vim, winnr():1, ft: vim, bt: )
BufFilePre: abuf: 12, amatch: …/vim/plugged/vim-gitgutter/meh.vim, winnr():1, ft: vim, bt: )
[...]

PR #622, opened 2019-06-03
Rewrite cached filepath at BufWritePost

pylipp:
resolves airblade/vim-gitgutter#551

### DIFF
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -207,6 +207,10 @@ augroup gitgutter
   autocmd CursorHold,CursorHoldI * call gitgutter#process_buffer(bufnr(''), 0)
   autocmd FileChangedShellPost   * call gitgutter#process_buffer(bufnr(''), 1)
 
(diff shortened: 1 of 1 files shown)


plugin/gitgutter.vim
```diff
@@ -207,6 +207,10 @@ augroup gitgutter
   autocmd CursorHold,CursorHoldI * call gitgutter#process_buffer(bufnr(''), 0)
   autocmd FileChangedShellPost   * call gitgutter#process_buffer(bufnr(''), 1)
 
+  " Reset cached filepath when e.g. executing :saveas on a tracked file
+  " autocmd BufWritePost * call s:on_bufwritepost()
+  autocmd BufWritePost * let b:gitgutter.path = expand('%:p')
```
airblade: This will only work if the current directory is the project root directory, i.e. not all the time.

airblade: You could try `unlet b:gitgutter.path`.  (Don't have time to test it right now...)

airblade:
Fixed in 07b60b5.

Issue #623, opened 2019-06-03
Undefined highlight attributes return invalid values from get_hl() in vim 7.4


jbylsma:
> What is the latest commit SHA in your installed vim-gitgutter?
https://github.com/airblade/vim-gitgutter/commit/a4e8d3f
> What vim/nvim version are you on?
7.4.629
> If no signs are showing at all, what does `:echo b:gitgutter.path` give?
N/A
> If no signs are showing at all, and the `path` value is a path and not `-2`, does it work with `let g:gitgutter_grep=''`?
N/A

With the latest master and a clean vimrc, Vim 7.4 has been throwing an error on startup: 
```
Error detected while processing function gitgutter#highlight#define_highlights:
line    6:
E421: Color name or number not recognized: ctermfg=-1 ctermbg=-1
Press ENTER or type command to continue
```

I Git bisected the issue back to https://github.com/airblade/vim-gitgutter/commit/9f21e3d, which is the refactor to simplify getting fg/bg highlights. It looks like older Vim versions would return -1 and not empty string from `synIDattr()` if the highlight was not defined; I can verify that empty string is being returned for vim 8.1.1450. I looked at vim's changelog but didn't see anything obvious about when that was switched. This wasn't an issue with the previous code because the regex parsing against `highlight` just wouldn't match.

As a workaround, I simply expanded the subsequent conditional to check for both empty string and -1. That ends up returning `NONE` back to other gitgutter functions and seems to be working fine. I'll push that patch as a PR.

airblade:
Thanks very much for looking into this and writing it all up.

PR #624, opened 2019-06-03
In get_hl(), discard returned synIDattr values of -1

jbylsma:
Addresses #623.

### DIFF
--- a/autoload/gitgutter/highlight.vim
+++ b/autoload/gitgutter/highlight.vim
@@ -133,7 +133,7 @@ endfunction
 
 function! s:get_hl(group, what, mode) abort
   let r = synIDattr(synIDtrans(hlID(a:group)), a:what, a:mode)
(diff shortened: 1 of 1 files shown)


autoload/gitgutter/highlight.vim
```diff
@@ -133,7 +133,7 @@ endfunction
 
 function! s:get_hl(group, what, mode) abort
   let r = synIDattr(synIDtrans(hlID(a:group)), a:what, a:mode)
-  if r ==# ''
+  if index(['', '-1'], r) >= 0
```
blueyed: I'd rather use `if empty(r) || r == -1` instead - no extra list.

jbylsma: Updated. I don't think I even grepped outside of the file for code styles, sorry about that.

airblade:
Thank you!

Issue #643, opened 2019-08-17
[Feature request] Hunk preview in a floating window


aymanbagabas:
Hi,

I just had this idea of having git hunks in a floating window instead of a preview window. Preview windows are so annoying because they push/move text in the current window.

So, I was wondering if this would be an appealing feature especially that Vim8 soon going to have floating windows and NeoVim already have floating windows in 0.4.

Thank you.

airblade:
I would be happy to incorporate floating windows as long as the user can choose not to use them.

> Preview windows are so annoying because they push/move text in the current window.

They don't for me.  My preview window opens below my other windows so it doesn't move the text at all.

ccjmne:
For the record, `splitkeep=screen` will prevent existing content from being pushed when opening a preview window!

PR #644, opened 2019-08-18
Add neovim hunk preview floating window

aymanbagabas:
This uses floating windows introduced in neovim to display hunks instead of
preview windows. Requires nvim with floating windows.

If nvim is running and has floating window capabilities, display hunk
preview in a floating window instead of a preview windows. Then a user
can focus to that floating window with Ctrl-w to be able to scroll if
needed to. Moving the cursor in the current buffer closes the floating
window or `:q` inside the floating window.

#643 

render1566093464359

I'm also using this in my config to override `diff` filetype background color.
```
highlight DiffAdded ctermbg=none
highlight DiffRemoved ctermbg=none
```

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -269,6 +269,7 @@ You can customise:
 * Whether vim-gitgutter runs asynchronously (defaults to yes)
 * Whether to clobber or preserve non-gitgutter signs
 * The priority of gitgutter's signs.
+* Whether vim-gitgutter should use floating windows for hunk preview (defaults to no)
 
 Please note that vim-gitgutter won't override any colours or highlights you've set in your colorscheme.
 
@@ -452,6 +453,11 @@ By default diffs are run asynchronously.  To run diffs synchronously instead:
 ```viml
 let g:gitgutter_async = 0
 ```
+
+#### To use floating windows for hunk preview
+
+By default hunk preview uses preview window. Make sure floating windows is supported.
+Add `let g:gitgutter_preview_win_floating = 1` to your `~/.vimrc`
 
 
 ### Extensions
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -300,7 +300,7 @@ function! s:preview(hunk_diff)
   let body_length = len(body)
   let previewheight = min([body_length, &previewheight])
 
-  if has('nvim') && exists('*nvim_open_win')
+  if g:gitgutter_preview_win_floating && has('nvim') && exists('*nvim_open_win')
     let buf = nvim_create_buf(v:false, v:false)
     let width = max(map(copy(body), 'strdisplaywidth(v:val)'))
     let height = body_length + (width / &columns)
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -300,22 +300,47 @@ function! s:preview(hunk_diff)
   let body_length = len(body)
   let previewheight = min([body_length, &previewheight])
 
-  silent! wincmd P
-  if !&previewwindow
-    noautocmd execute g:gitgutter_preview_win_location previewheight 'new'
-    set previewwindow
+  if has('nvim') && exists('*nvim_open_win')
+    let buf = nvim_create_buf(v:false, v:false)
+    let width = max(map(copy(lines), 'strdisplaywidth(v:val)'))
+    let height = len(lines)
+
+    let s:floating_win = nvim_open_win(
+          \ buf, v:true, {
+          \   'width': width,
+          \   'height': height,
+          \   'relative': 'cursor',
+          \   'anchor': 'NW',
+          \   'row': 1,
+          \   'col': -1
+          \ })
+
+    setlocal nonumber norelativenumber signcolumn=no nofoldenable
   else
-    execute 'resize' previewheight
+    silent! wincmd P
+    if !&previewwindow
+      noautocmd execute g:gitgutter_preview_win_location previewheight 'new'
+      set previewwindow
+    else
+      execute 'resize' previewheight
+    endif
   endif
 
   let b:hunk_header = header
 
   setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
+  setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile nospell
   execute "%delete_"
   call setline(1, body)
   normal! gg
 
   noautocmd wincmd p
+  augroup GitGutterFloatingWindow
+    autocmd!
+    autocmd CursorMoved <buffer> if exists('s:floating_win') |
+          \ let winnr = win_id2win(s:floating_win) |
+          \ if winnr > 0 | execute winnr.'wincmd c' | endif | endif
+  augroup END
 endfunction
 
 
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -302,8 +302,8 @@ function! s:preview(hunk_diff)
 
   if has('nvim') && exists('*nvim_open_win')
     let buf = nvim_create_buf(v:false, v:false)
-    let width = max(map(copy(lines), 'strdisplaywidth(v:val)'))
-    let height = len(lines)
+    let width = max(map(copy(body), 'strdisplaywidth(v:val)'))
+    let height = body_length + (width / &columns)
 
     let s:floating_win = nvim_open_win(
           \ buf, v:true, {
@@ -328,7 +328,6 @@ function! s:preview(hunk_diff)
 
   let b:hunk_header = header
 
-  setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
   setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile nospell
   execute "%delete_"
   call setline(1, body)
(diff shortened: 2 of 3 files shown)


airblade:
How do I move my cursor into the floating window?

autoload/gitgutter/hunk.vim
```diff
@@ -300,22 +300,47 @@ function! s:preview(hunk_diff)
...
   let previewheight = min([body_length, &previewheight])
 
-  silent! wincmd P
-  if !&previewwindow
-    noautocmd execute g:gitgutter_preview_win_location previewheight 'new'
-    set previewwindow
+  if has('nvim') && exists('*nvim_open_win')
+    let buf = nvim_create_buf(v:false, v:false)
+    let width = max(map(copy(lines), 'strdisplaywidth(v:val)'))
+    let height = len(lines)
```
airblade: The height has already been calculated above: `body_length`.

autoload/gitgutter/hunk.vim
```diff
@@ -300,22 +300,47 @@ function! s:preview(hunk_diff)
   let body_length = len(body)
   let previewheight = min([body_length, &previewheight])
 
-  silent! wincmd P
-  if !&previewwindow
-    noautocmd execute g:gitgutter_preview_win_location previewheight 'new'
-    set previewwindow
+  if has('nvim') && exists('*nvim_open_win')
+    let buf = nvim_create_buf(v:false, v:false)
+    let width = max(map(copy(lines), 'strdisplaywidth(v:val)'))
```
airblade: You need `body` instead of `lines`.

autoload/gitgutter/hunk.vim
```diff
@@ -300,22 +300,47 @@ function! s:preview(hunk_diff)
...
+      set previewwindow
+    else
+      execute 'resize' previewheight
+    endif
   endif
 
   let b:hunk_header = header
 
   setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
+  setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile nospell
```
airblade: Why duplicate the line above just to add a keyword?

autoload/gitgutter/hunk.vim
```diff
@@ -300,22 +300,47 @@ function! s:preview(hunk_diff)
...
-  if !&previewwindow
-    noautocmd execute g:gitgutter_preview_win_location previewheight 'new'
-    set previewwindow
+  if has('nvim') && exists('*nvim_open_win')
+    let buf = nvim_create_buf(v:false, v:false)
+    let width = max(map(copy(lines), 'strdisplaywidth(v:val)'))
+    let height = len(lines)
+
+    let s:floating_win = nvim_open_win(
+          \ buf, v:true, {
```
airblade: The `v:true` doesn't seem to work for me, i.e. my cursor doesn't enter the floating window.

autoload/gitgutter/hunk.vim
aymanbagabas: Accidentally added when rebasing, I was working with an older commit.

autoload/gitgutter/hunk.vim
aymanbagabas: It does, and that's how `setlocal` works in the floating window. `noautocmd wincmd p` gets you back to the previous window. Checkout `:h api-floatwin`

autoload/gitgutter/hunk.vim
aymanbagabas: Right, I should use that. I was thinking of adding height offset in case width is greater than `&columns` so that the floating window would display wrapped lines.

blueyed:
This looks nice!

`:GitGutterStageHunk` does not appear to work from the floating window?

(btw: I've wondered if it would make sense to have a mode where it gets enabled to be triggered on CursorHold always (i.e. when moving around)? - can be done manually though easily I guess)

blueyed:
> `:GitGutterStageHunk` does not appear to work from the floating window?

Maybe `:w` should stage it from there?  (so that `:wq` would write and close it after you might have edited it)

airblade:
>  I've wondered if it would make sense to have a mode where it gets enabled to be triggered on CursorHold always

What's "it" here?  The hunk preview?

> Maybe :w should stage it from there? (so that :wq would write and close it after you might have edited it)

Sounds good.  I suppose we could also have `:w` in a preview window stage the contents.

autoload/gitgutter/hunk.vim
airblade: My cursor does not enter the floating window.  When I invoke a hunk preview the cursor stays where it is and the floating window is displayed below it.  If I move the cursor the floating window disappears (as expected when the cursor is in the original window).

blueyed: That's due to moving out of it later: https://github.com/aymanbagabas/vim-gitgutter/blob/3c75c97c3dfa12f0e868b13ed9344071b54a1d0a/autoload/gitgutter/hunk.vim#L336

blueyed:
> What's "it" here? The hunk preview?

Yes.
It would not close the floating window / preview window, but update it for new hunks when moving around.
Not really useful by default, but maybe a separate command?
Thought about it when seeing https://github.com/rhysd/git-messenger.vim.

> I suppose we could also have `:w` in a preview window stage the contents.

Cool.  :+1: for consistency.

autoload/gitgutter/hunk.vim
blueyed: btw: it would be better to not use "noautocmd" there, for plugins tracking the current window etc.  At least nvim will trigger the autocmds for the floating window, but not when going back then.

airblade: > That's due to moving out of it later

Ah, of course.  Thanks.

airblade: > it would be better to not use "noautocmd" there

That's been there since the feature was introduced.  I can't remember now why I didn't want autocmds.

airblade:
> I suppose we could also have :w in a preview window stage the contents.

Done in 77fb3fd.

aymanbagabas:
> (btw: I've wondered if it would make sense to have a mode where it gets enabled to be triggered on CursorHold always (i.e. when moving around)? - can be done manually though easily I guess)

I like that idea, a GitGutter mode. I was experimenting with something similar using CursosMoved but ended up not using it because it caused a lot of lag. Could be improved though.

```
" auto close gitgutter preview window and update it when next line is a hunk.
autocmd CursorMoved * if exists('g:__gitgutter_hunk_preview') && g:__gitgutter_hunk_preview |
\   if gitgutter#hunk#in_hunk(line('.')) |
  \     call gitgutter#hunk#preview() |
  \   else |
  \     silent! wincmd P |
  \     if &previewwindow | pclose | silent! wincmd p | endif |
  \     let g:__gitgutter_hunk_preview=0 |
\   endif |
\ endif
```
Then i had <leader>hp sets `g:__gitgutter_hunk_preview=1` and moving with `]h`

blueyed:
btw: I think it would be good to keep the focus in the preview/floating window - so that you can directly edit and/or write it to stage it.

blueyed:
@aymanbagabas btw: https://github.com/airblade/vim-gitgutter/pull/645 (in case you have not seen it)

aymanbagabas:
> @aymanbagabas btw: #645 (in case you have not seen it)

Tested and seems to work! Should I rebase this or wait until #645 gets pulled?

blueyed:
I'd wait for it to get pulled - might change still etc, and better to keep it separate.
It is not really required here.

README.mkd
```diff
@@ -269,6 +269,7 @@ You can customise:
 * Whether vim-gitgutter runs asynchronously (defaults to yes)
 * Whether to clobber or preserve non-gitgutter signs
 * The priority of gitgutter's signs.
+* Whether vim-gitgutter should use floating windows for hunk preview (defaults to no)
```
blueyed:
```suggestion
* Whether vim-gitgutter should use floating windows for hunk preview (defaults to no).
```

README.mkd
```diff
@@ -453,6 +454,11 @@ By default diffs are run asynchronously.  To run diffs synchronously instead:
 let g:gitgutter_async = 0
 ```
 
+#### To use floating windows for hunk preview
+
+By default hunk preview uses preview window. Make sure floating windows is supported.
+Add `let g:gitgutter_preview_win_floating = 1` to your `~/.vimrc`
```
blueyed: This needs to be rephrased.

plugin/gitgutter.vim
```diff
@@ -22,6 +22,7 @@ function! s:set(var, default) abort
   endif
 endfunction
 
+call s:set('g:gitgutter_preview_win_floating',        0)
```
blueyed: It should probably default to use floating windows if supported.

airblade: I'm happy with that.

aymanbagabas: Sorry I'm confused, should this be true or false?

airblade:
Thanks for all the discussion here.  I implemented it (with changes) in e929cb8...67c8cca.

vilari-mickopf:
> > How do I move my cursor into the floating window?
> 
> Ctrl-w w would move you to the floating window.

This is bad, since it won't work when you have split windows, and people tend to have split windows all the time. Because of that Ctrl-w p should be used, but in this case it's not working.

PR #645, opened 2019-08-21
[WIP] s:preview: use BufWriteCmd

blueyed:
Ref: https://github.com/airblade/vim-gitgutter/commit/77fb3fd81e78afd96ef017b2ad9107c1775a78b5#r34782119

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -310,14 +310,16 @@ function! s:preview(hunk_diff)
 
   let b:hunk_header = header
 
-  setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
+  setlocal noreadonly modifiable filetype=diff bufhidden=delete noswapfile
   execute "%delete_"
   call setline(1, body)
   normal! gg
 
-  cnoreabbrev <buffer> <expr> w  getcmdtype() == ':' && getcmdline() == 'w'  ? 'GitGutterStageHunk' : 'w'
-  " Staging hunk from the preview window closes the window anyway.
-  cnoreabbrev <buffer> <expr> wq getcmdtype() == ':' && getcmdline() == 'wq' ? 'GitGutterStageHunk' : 'wq'
+  exe 'file gitgutter://hunk-'.bufnr('%')
+  set nomodified
+  augroup gitgutter_stagehunk
+    autocmd! BufWriteCmd <buffer> set nomodified | GitGutterStageHunk
+  augroup END
 
   noautocmd wincmd p
 endfunction
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -192,7 +192,7 @@ endfunction
 function! s:hunk_op(op, ...)
   let bufnr = bufnr('')
 
-  if &previewwindow
+  if exists('b:_gitgutter_preview_for')
     if string(a:op) =~ '_stage'
       " combine hunk-body in preview window with updated hunk-header
       let hunk_body = getline(1, '$')
@@ -214,9 +214,7 @@ function! s:hunk_op(op, ...)
 
       let hunk_diff = join(hunk_header + hunk_body, "\n")."\n"
 
-      wincmd p
-      pclose
-      call s:stage(hunk_diff)
+      call s:stage(hunk_diff, b:_gitgutter_preview_for)
     endif
 
     return
@@ -257,8 +255,8 @@ function! s:hunk_op(op, ...)
 endfunction
 
 
-function! s:stage(hunk_diff)
-  let bufnr = bufnr('')
+function! s:stage(hunk_diff, ...) abort
+  let bufnr = a:0 ? a:1 : bufnr('')
   let diff = s:adjust_header(bufnr, a:hunk_diff)
   " Apply patch to index.
   call gitgutter#utility#system(
@@ -293,6 +291,8 @@ endfunction
 
 
 function! s:preview(hunk_diff)
+  let bufnr = bufnr('%')
+
   let lines = split(a:hunk_diff, '\n')
   let header = lines[0:4]
   let body = lines[5:]
@@ -315,7 +315,8 @@ function! s:preview(hunk_diff)
   call setline(1, body)
   normal! gg
 
-  exe 'file gitgutter://hunk-'.bufnr('%')
+  let b:_gitgutter_preview_for = bufnr
+  exe 'file gitgutter://hunk-'.b:_gitgutter_preview_for
   set nomodified
   augroup gitgutter_stagehunk
     autocmd! BufWriteCmd <buffer> set nomodified | GitGutterStageHunk


autoload/gitgutter/hunk.vim
```diff
@@ -310,14 +310,18 @@ function! s:preview(hunk_diff)
...
-  setlocal noreadonly modifiable filetype=diff buftype=nofile bufhidden=delete noswapfile
+  setlocal noreadonly modifiable filetype=diff bufhidden=delete noswapfile
   execute "%delete_"
   call setline(1, body)
   normal! gg
 
-  cnoreabbrev <buffer> <expr> w  getcmdtype() == ':' && getcmdline() == 'w'  ? 'GitGutterStageHunk' : 'w'
-  " Staging hunk from the preview window closes the window anyway.
-  cnoreabbrev <buffer> <expr> wq getcmdtype() == ':' && getcmdline() == 'wq' ? 'GitGutterStageHunk' : 'wq'
+  exe 'file gitgutter://hunk-'.bufnr('%')
```
blueyed: Given the filename that it has now, this could be checked in https://github.com/blueyed/vim-gitgutter/blob/944181f1528887051daf37a0781ad7c2aac97634/autoload/gitgutter/hunk.vim#L195 (instead of "if &previewwindow"), to handle the upcoming nvim floating window there also maybe.

airblade:
I think this approach is conceptually neater.  However there are a couple of hitches:

When you stage a hunk from the preview window using `:GitGutterStageHunk` (or the equivalent mapping), the plugin stages the hunk and closes the preview window.  This makes sense to me.

Normally `:wq` writes and closes the current window, so a user might be surprised when it closes both the preview window (as expected) and also the original window (unexpected).

I think it would be good to have a way to make `:w` and `:wq` behave the same in your implementation – would you agree?

Related to this, if you preview a hunk, go to the preview window, and amend the hunk e.g. by deleting a line, you can no longer close the window without writing the buffer.  Perhaps this is actually better than the current implementation but I'm not sure.

blueyed:
> Normally `:wq` writes and closes the current window, so a user might be surprised when it closes both the preview window (as expected) and also the original window (unexpected).

Sure, that's a bug.  It's because of:  https://github.com/airblade/vim-gitgutter/blob/1d142daeb7d71bfedd3efd8c95d44c51aa73dcc8/autoload/gitgutter/hunk.vim#L217-L218

Just skipping this does not work, since `s:stage` expects to be in the buffer.  That could be fixed by passing through buffer numbers etc, but there seems to be a more fundamental issue with it, e.g. with "wincmd p" would also break / not work as expected in case you've gone to another window than the one where the preview was opened from inbetween.

As stated elsewhere, I think it would be good to have the focus in the preview window by default (so you can quit/write it directly), but I can also see that it might act for preview really only - especially with a possible CursorMoved-mode of course.

autoload/gitgutter/hunk.vim
blueyed: Used an explicit buffer-local variable for this now.  Easier to than matching the bufnr from the bufname.. ;)

autoload/gitgutter/hunk.vim
```diff
@@ -192,7 +192,7 @@ endfunction
 function! s:hunk_op(op, ...)
   let bufnr = bufnr('')
 
-  if &previewwindow
+  if exists('b:_gitgutter_preview_for')
```
blueyed: This should then also work with floating windows (#644).

airblade:
I want the preview window to close automatically when I stage something from it; at that point it's out of date and there's no reason for it to remain open.  That's why the code does that :)

> "wincmd p" would also break / not work as expected in case you've gone to another window...

It makes sense to me that you return to the window you were in before the preview window, even if it's not the one from which the user previewed the hunk.

> I think it would be good to have the focus in the preview window by default...

I preview hunks (and then close the preview) far more often than I stage hunks from the preview window.  So for me it makes sense to leave the cursor in the original window; the user can move to the preview window if they want.

blueyed:
> I want the preview window to close automatically when I stage something from it; at that point it's out of date and there's no reason for it to remain open. That's why the code does that :)

That's a common use case I guess.. :)
Do you typically then stage all of it or only certain hunks also?  (not sure if that works, have not used it much myself)

> > "wincmd p" would also break / not work as expected in case you've gone to another window...
> 
> It makes sense to me that you return to the window you were in before the preview window, even if it's not the one from which the user previewed the hunk.

True.  But I've meant that 's:stage()` then gets called with it, and takes that bufnr (I've changed this here already).  But it means that currently the buffer might be off, and then you get an error that it could not be staged.



> > I think it would be good to have the focus in the preview window by
> > default...
> 
> I preview hunks (and then close the preview) far more often than I stage
> hunks from the preview window. So for me it makes sense to leave the cursor
> in the original window; the user can move to the preview window if they want.

If you close the preview usually anyway (manually), you could also close it
from the preview window itself - which could have a short mapping for it,
like `Q` to close it (but `:q<cr>` might be good enough anyway, and I for
example have a `Q` map for this anyway (i.e. let users handle it)).

I can see though that it is good to open the preview, not focus it, and have
it closed when the cursor is moved automatically.

airblade:
> Do you typically then stage all of it or only certain hunks also?

Mostly I just want to remind myself what the code looked like before I changed it. When it comes to staging, I usually stage the whole hunk (from the original window); only occasionally do I edit the hunk in the preview window and then stage it.

> But it means that currently the buffer might be off...

I still don't get this ;)  I'll try to figure it out.

> If you close the preview usually anyway (manually), you could also close it
from the preview window itself...

That is a good point.  I think this is more about "flow" than keystrokes: whether or not the cursor moves to the preview window I need to type something to close it; but it feels like an interruption if the cursor moves.

> I can see though that it is good to open the preview, not focus it, and have
it closed when the cursor is moved automatically.

Yes, though sometimes I like to keep the preview open for a while :)

I think this is tricky because there isn't a single workflow.  Sometimes I want the preview to open and close automatically; other times I want it to be manual; sometimes I want the preview to remain open; sometimes I want to go to the preview and edit and stage; sometimes I want to go to the preview and edit but then discard; etc.  These are contradictory so I suppose the best thing is to (try to) identify the common case and make that the default, and allow the user to vary the behaviour either by configuration or nice easy commands/mappings.

airblade:
| Action | Possibilities (* current default) |
|--|--|
| Open hunk preview | automatic \| manual* |
| Close hunk preview | automatic \| manual* |
| Focus the preview | yes \| no* |

And staging from the preview always closes the preview.

Generally when writing software I try to avoid adding options but vim plugins seem to be an exception ;)  Do you think making the default behaviour configurable would be a good idea?  We could then choose whether to provide commands and even maps to change the defaults, or just leave that to the user.

airblade:
This is my latest patch based on 2980a16:

```diff
diff --git i/autoload/gitgutter/hunk.vim w/autoload/gitgutter/hunk.vim
index 16befc9..89cee8d 100644
--- i/autoload/gitgutter/hunk.vim
+++ w/autoload/gitgutter/hunk.vim
@@ -403,9 +403,10 @@ function! s:open_hunk_preview_window()
             \ 'style': 'minimal'
             \ })
       call nvim_buf_set_option(buf, 'filetype',  'diff')
-      call nvim_buf_set_option(buf, 'buftype',   'nofile')
+      call nvim_buf_set_option(buf, 'buftype',   'acwrite')
       call nvim_buf_set_option(buf, 'bufhidden', 'delete')
       call nvim_buf_set_option(buf, 'swapfile',  v:false)
+      call nvim_buf_set_name(buf, 'gitgutter://hunk-preview')
 
       " Assumes cursor is in original window.
       autocmd CursorMoved <buffer> ++once call s:close_hunk_preview_window()
@@ -429,9 +430,9 @@ function! s:open_hunk_preview_window()
 
   silent! wincmd P
   if !&previewwindow
-    noautocmd execute g:gitgutter_preview_win_location &previewheight 'new'
+    noautocmd execute g:gitgutter_preview_win_location &previewheight 'new gitgutter://hunk-preview'
     set previewwindow
-    setlocal filetype=diff buftype=nofile bufhidden=delete
+    setlocal filetype=diff buftype=acwrite bufhidden=delete
     " Reset some defaults in case someone else has changed them.
     setlocal noreadonly modifiable noswapfile
   endif
@@ -473,23 +474,9 @@ function! s:populate_hunk_preview_window(header, body)
 endfunction
 
 
-" Floating window: does not care where cursor is.
-" Preview window: assumes cursor is in preview window.
 function! s:enable_staging_from_hunk_preview_window()
-  if g:gitgutter_preview_win_floating
-    " Move cursor to previewing window without triggering autocmd which closes it.
-    " This is necessary because there is no way to set a cabbrev in an arbitrary buffer.
-    execute 'noautocmd' win_id2win(s:winid).'wincmd w'
-  endif
-
-  cnoreabbrev <buffer> <expr> w  getcmdtype() == ':' && getcmdline() == 'w'  ? 'GitGutterStageHunk' : 'w'
-  " Staging hunk from the previewing window closes the window anyway.
-  cnoreabbrev <buffer> <expr> wq getcmdtype() == ':' && getcmdline() == 'wq' ? 'GitGutterStageHunk' : 'wq'
-
-  if g:gitgutter_preview_win_floating
-    " Move cursor back without triggering autocmd.
-    noautocmd wincmd p
-  endif
+  execute 'autocmd BufWriteCmd <buffer='.winbufnr(s:winid).'> set nomodified | GitGutterStageHunk'
+  execute 'autocmd QuitPre <buffer='.winbufnr(s:winid).'> set nomodified'
 endfunction
 ```

However there's still a problem with the Vim warning the user about discarding unsaved changes in the preview window.  I want the preview window to be a scratch buffer so that Vim never complains about discarding its unsaved changes.

If I open a file with changes, preview a hunk, go to the preview window, delete a line, then go back to the file, then `:q`, Vim complains about the preview window.

blueyed:
> However there's still a problem with the Vim warning the user about discarding unsaved changes in the preview window. I want the preview window to be a scratch buffer so that Vim never complains about discarding its unsaved changes.

I think it is good to see a warning after you've done changes, and not saving them.  After all you've done changes, and therefore should confirm that you do not want to apply them then.
(There should be no warning when doing no changes of course.)

Feel free to force-push your patch here - or create a new PR from it.

airblade:
> I think it is good to see a warning after you've done changes, and not saving them.

You're right, that does make sense.

I've created a new PR at #647 which works for me.  Does it do what you need?

PR #646, opened 2019-08-21
test/test: minor improvements

blueyed:


test/test
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
airblade: This prevents `messages.log` from being removed later on.

airblade:
I applied the other change in 85a1bf7.  Thanks.

Issue #653, opened 2019-09-21
Preview in floating window: error before display


JoshMcguigan:
Before I start, I'd like to thank you for your work on this plugin. I've really enjoyed it and I was really happy to see you recently added an option to show the preview in a floating window. 

I've enabled the feature with `let g:gitgutter_preview_win_floating = 1` in $MYVIMRC, and it works great, as long as I havn't used netrw in that vim session. If I have used netrw, I get several error messages when I try to display vim-gitgutter preview (using <leader>hp). 

```
Error detected while processing /usr/share/vim/vim81/syntax/netrw.vim:                                                                                                                    
line  128:
E930: Cannot use :redir inside execute()
line  130:
E930: Cannot use :redir inside execute()
Press ENTER or type command to continue
```

After pressing enter here, the preview popup works as expected.

> What is the latest commit SHA in your installed vim-gitgutter?
76aa5920dba9b05991940f7dc29dc79ad7af0a3e
> What vim/nvim version are you on?
vim 8.1.1890

Let me know if I can provide any other diagnostic information.

Thanks!

blueyed:
For refererence, the netrw code that triggers it: https://github.com/vim/vim/blob/9a4a8c4d5993c6371486c895a515c2ad351e9aaa/runtime/syntax/netrw.vim#L128-L130

The first question is: why does "syntax netrw" get used, and where?
Does the error show up before the floating window already?

gitgutter appears to trigger it via https://github.com/airblade/vim-gitgutter/blob/76aa5920dba9b05991940f7dc29dc79ad7af0a3e/autoload/gitgutter/hunk.vim#L425 - try commenting that line.

Also try the following patch instead:
```
diff --git i/autoload/gitgutter/hunk.vim w/autoload/gitgutter/hunk.vim
index d196d68..becddfc 100644
--- i/autoload/gitgutter/hunk.vim
+++ w/autoload/gitgutter/hunk.vim
@@ -422,7 +422,7 @@ function! s:open_hunk_preview_window()
             \ })
 
       call setbufvar(winbufnr(s:winid), '&filetype', 'diff')
-      call win_execute(s:winid, 'syntax enable')
+      call win_execute(s:winid, 'doautocmd FileType')
 
       return
     endif
```

JoshMcguigan:
Wow, thanks for the quick response. Either solution (commenting line 425, and replacing it as shown in the patch) solves the problem.

blueyed:
Cool.
[...]

PR #654, opened 2019-09-21
Fix preview-floatwin with Vim: use doautocmd FileType

blueyed:
Using "syntax enable" might trigger netrw for unknown reasons, resulting
in an error.

Fixes https://github.com/airblade/vim-gitgutter/issues/653.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -422,7 +422,7 @@ function! s:open_hunk_preview_window()
             \ })
 
       call setbufvar(winbufnr(s:winid), '&filetype', 'diff')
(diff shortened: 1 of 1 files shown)


autoload/gitgutter/hunk.vim
```diff
@@ -422,7 +422,7 @@ function! s:open_hunk_preview_window()
             \ })
 
       call setbufvar(winbufnr(s:winid), '&filetype', 'diff')
-      call win_execute(s:winid, 'syntax enable')
+      call win_execute(s:winid, 'doautocmd FileType')
```
blueyed: `win_execute` might not be needed altogether: https://github.com/vim/vim/issues/4961

PR #661, opened 2019-10-23
Fix example mappings

localstatic:
I'm guessing this change was made to demonstrate the functionality shown in the screenshot (this change was introduced at the same time the screenshot and associated description was updated, in commit 6660aca9476b7ddd209cbda0329a59904eda53ed), but since these mappings don't actually work (and indeed emit warnings), I'm guessing that change was inadvertently committed.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -142,11 +142,9 @@ Both of those take a preceding count.
 
 To set your own mappings for these, for example `]h` and `[h`:
 
-In the lines below I have deleted brackets around GitGutter{Next,Prev}Hunk:
-
 ```viml
-nmap ]h <Plug>GitGutterNextHunk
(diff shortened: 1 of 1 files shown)


README.mkd
```diff
@@ -142,11 +142,9 @@ Both of those take a preceding count.
 
 To set your own mappings for these, for example `]h` and `[h`:
 
-In the lines below I have deleted brackets around GitGutter{Next,Prev}Hunk:
```
blueyed: Werid comment?!

airblade: As @localstatic guessed, I accidentally committed the edits I made to just to have some changes to show in a screenshot.

airblade:
@localstatic Thanks!

PR #664, opened 2019-11-16
Add g:gitgutter_use_location_list option

rcr:
Thought it would be useful to others to have this fix that I wrote for myself. I use the global quickfix list to hold compiler errors, which I don't want to have clobbered by other things such as `:grep` and `:GitGutterQuickFix`. Also, I think semantically, this makes more sense anyways, these are locations in the code, not fixable items.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -18,7 +18,7 @@ Features:
 * Diffs against index (default) or any commit.
 * Allows folding all unchanged text.
 * Provides fold text showing whether folded lines have been changed.
-* Can load all hunk locations into quickfix list.
+* Can load all hunk locations into quickfix list or the current window's location list.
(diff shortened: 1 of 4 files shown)


airblade:
Thanks for this.

I went with the quickfix list because the code locations could be in any file, not just the buffer in the current window – so to me it makes more sense to use a global list.  However I appreciate your point and am happy to merge this.

Please could you update the readme as well?

rcr:
Ah. That makes sense. I use a single window and manage multiple buffers, so its just a difference in workflow.

I updated the readme.

Thanks for your hard work on this wonderful plugin!

airblade:
Thank you very much!

Issue #599, opened 2019-03-20
Startup is very slow


DroneKid:
Hi!

The startup is very slow on the latest master. It is blazingly fast if I roll back to `932ffaca092cca246b82c33e23d2d3a05e192e08`. It looks like `5bfe5b92099e0d332a7ca26c0263fab509ee9617 Big refactor` had introduced something to slow it down.

Slow startup is only experrienced if I pass a file to vim at command line.

> What is the latest commit SHA in your installed vim-gitgutter?
`7eeea63e62b1cc088a75c7a7c244fc774d82e5bb`

> What vim/nvim version are you on?
vim 8.1.1000

Cheers,
A.

airblade:
I can't reproduce this.  Please could you [profile](https://github.com/airblade/vim-gitgutter/wiki/How-to-profile-vim-gitgutter) it?

Morozzzko:
I'm having a similar problem. 

I'm at `7eeea63e62b1cc088a75c7a7c244fc774d82e5bb`
NVIM v0.3.4

I'm going to attach the problematic parts of the log:

```
SCRIPT  /Users/moroz/.vim/plugged/vim-gitgutter/autoload/gitgutter/diff.vim
Sourced 1 time
Total time:   3.610975
 Self time:   0.000804
```

Here's the gist with profiling for `diff.vim`

https://gist.github.com/Morozzzko/5a1792f2cbc3ef0e12c845c66bc15e9e#file-gistfile1-txt-L17

blueyed:
@Morozzzko 
So it seems the git call itself is slow (https://github.com/nedbat/coveragepy/blob/820b255f34a0aac8670b0c819153bb8b38c4b2c6/coverage/ctracer/tracer.c#L1015-L1017).  Can you run it manually?  Is it due to your shell startup being slow maybe etc?

Morozzzko:
Right, thank you! That's what happened. Shell startup was slow, so I optimized it and everything became better. Thank you!

Sorry for hijacking the issue by the way. Seems like I've got a different kind of problem

nicksergeant:
I'm also seeing about a .5s delay upon opening any new file into a new buffer, and I've narrowed it down to `vim-gitgutter`. I'm using NERDtree to open a file, and the problem only exists for new buffers. Once a file has been opened in the current Vim session, opening the same file is very fast (under 100ms, I'd guess).

[...]

Issue #531, opened 2018-06-19
Loading takes 4.5x longer with FISH vs BASH


LukeChannings:
With the default shell set to fish, vim-gitgutter loads very slowly.

Profile excerpt with `set shell=fish`:

```
223.547  121.853: opening buffers
224.394  000.404  000.404: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter.vim
225.680  000.236  000.236: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/async.vim
226.197  000.255  000.255: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/debug.vim
247.461  019.762  019.762: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/diff.vim
```

and without:

```
211.675  108.961: opening buffers
212.575  000.454  000.454: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter.vim
213.884  000.245  000.245: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/async.vim
214.450  000.330  000.330: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/debug.vim
402.039  185.952  185.952: sourcing /Users/44092098/.local/share/nvim/plugged/vim-gitgutter/autoload/gitgutter/diff.vim
```

airblade:
Sourcing `diff.vim` executes [this line](https://github.com/airblade/vim-gitgutter/blob/a986ab054788776dca269d6c289b470255d54e8c/autoload/gitgutter/diff.vim#L7):

```viml
  call system(g:gitgutter_git_executable.' -c foo.bar=baz --version')
```

Presumably that is the cause of the slow down.

If you add `return 1` after that line, does the slowdown go away?

LukeChannings:
Nope, looking at the order of execution, it seems the slowdown is in `debug.vim`, `diff.vim` only takes ~10ms to run.

LukeChannings:
This is with Vim 8.1.1, fyi.

LukeChannings:
Oh, apologies, I stupidly did:

```vim
" True for git v1.7.2+.
function! s:git_supports_command_line_config_override() abort
[...]

PR #675, opened 2020-01-21
Use known shell for git_supports_command_line_config_override.

dewyatt:
This can speed things up significantly if your normal shell does a lot of init stuff.


airblade:
Thanks!  Using the normal `system()` was a bit of an oversight by me.

This might help with #531 and #599 too.

PR #676, opened 2020-02-11
Open quickfix window automatically

balki:
This makes it similar to default commands that open quickfix when called  - `make`, `grep`

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -209,7 +209,9 @@ function! gitgutter#quickfix()
   endfor
   if !g:gitgutter_use_location_list
     call setqflist(locations)
+    copen
   else
     call setloclist(0, locations)
+    lopen
   endif
 endfunction


airblade:
I don't think that `make` / `grep` / `vimgrep` opens the quickfix window.  In fact the help for `grep` has a note that:

> :copen (or :lopen for :lgrep) may be used to open a buffer containing the search results in linked form.

Which suggests that by default this doesn't happen.

balki:
You are right. Apologies. It is just my setup. However I think it is a good idea to have it open by default. Please close the PR if the current default is fine.

airblade:
I don't really have an opinion either way but in general I like to remain consistent with Vim, so I'll leave it as it is.

You could add something like this to your vimrc instead:

```viml
nmap <Leader>gq :GitGutterQuickFix\|copen<CR>
```

PR #677, opened 2020-02-13
Set scriptencoding for file with utf8 BOM

blueyed:
(reported by vint)


airblade:
That's a good catch.  Thank you!

Issue #614, opened 2019-05-21
Signs are color grey


plgrazon:
> What is the latest commit SHA in your installed vim-gitgutter?

commit e5936a22cfd2e25893e2ac34af4b0afdec1e36f5 (grafted, HEAD -> master, origin/master, origin/HEAD)
Author: Andy Stewart <boss@airbladesoftware.com>
Date:   Wed May 15 16:27:17 2019 +0100

Fix tmux command.

> What vim/nvim version are you on?

VIM - Vi IMproved 8.1 (2018 May 18, compiled Feb 13 2019 17:42:00)
macOS version
Included patches: 1-900
Compiled by Homebrew
 
Signs are visible but they are color grey. I tried modifying the colors like stated on the docs but they're not working. I'm using desert as my colorsheme and iterm2 for my terminal

![Screenshot 2019-05-21 13 56 49](https://user-images.githubusercontent.com/33631935/58130188-515d6500-7bd0-11e9-8846-ebb85b91a27d.png)

airblade:
This is probably due to changes made in faa1e95...3ce2a4b: previously the plugin set its own colour values for the signs; those changes made the plugin use colours from the colorscheme instead.  In theory this should work better across colorschemes but in practice it has caused a bit of trouble.

I'll try to reproduce this with your colorscheme...

plgrazon:
" Git Gutter"
 set updatetime=250
 let g:gitgutter_max_signs = 500
 " No mapping
 let g:gitgutter_map_keys = 0
 " Colors
 let g:gitgutter_override_sign_column_highlight = 0
 highlight clear SignColumn
 highlight GitGutterAdd    ctermfg=2
 highlight GitGutterChange ctermfg=3
 highlight GitGutterDelete ctermfg=1
 highlight GitGutterChangeDelete ctermfg=4

JPFrancoia:
This fixed my issue too but it'd be nice to know why we have to specify numbers instead of the colours names

airblade:
I don't know either why the colour names don't work.

adah1972:
Currently with HEAD I am getting black on dark grey with [my colour scheme](https://github.com/mbbill/desertEx). Commit faa1e95 has much better default colours.

My test shows changing `highlight GitGutterAdd…` in .vimrc has no effect. Doing it after a file is loaded is OK… Sad. Do I have to to modify the colour scheme files? It won’t work well on system default schemes, if I want to change the gitgutter colours for them.

[...]

PR #684, opened 2020-03-19
Implement an alternative solution to sign colors.

adah1972:
A fix to https://github.com/airblade/vim-gitgutter/issues/614.


airblade:
Thanks for this.

Please could you rename `g:gitgutter_use_scheme_diff_color` to `g:gitgutter_use_colorscheme`?

Also, if someone has defined the `GitGutterAdd` highlight group in their config, will this change still use it?

adah1972:
Sure. I will make the change.

Yes, their own `GitGutterAdd` will still work (with gotchas as I mentioned in the discussion, loading order and timing). In fact, I have such settings in my .vimrc, as I have shortcut keys to switch colour schemes...

airblade:
Also the documentation needs updating.

airblade:
On reflection it's not very vim-like to configure colours in variables with values like `guifg=#009900 ctermfg=2`.  The vim way is to configure highlight groups.

adah1972:
I think I misunderstood the meaning of the variable. You meant just the renaming of the variable, right?

airblade:
> Please could you rename `g:gitgutter_use_scheme_diff_color` to `g:gitgutter_use_colorscheme`

Yes, just changing the name.

adah1972:
I think my current way is easier, and we are not introducing new highlight groups: they are just default values. If I misunderstood you, please explain in more details.

airblade:
This makes sense to me:

- If `GitGutterAdd` etc is defined, use it.
- Else if `g:gitgutter_use_colorscheme` is 0 (default), define `GitGutterAdd` with the nice default colours.
- Else use the colorscheme's colours.

Would that suit you?

adah1972:
The problem with a user-defined `GitGutterAdd` is that they need to specify the background colours manually, which makes it less useful. But I can implement that logic.

adah1972:
With the current changes, I feel `g:gitgutter_use_colorscheme` might be a misnomer (the colour scheme's GitGutterAdd… always takes precendence), and `g:gitgutter_use_scheme_diff_color` may be better. What do you think?

adah1972:
Documentation is updated. Let me know if you want to rename `g:gitgutter_use_colorscheme` back to `g:gitgutter_use_scheme_diff_color`. I feel like that better.

README.mkd
```diff
@@ -308,7 +308,11 @@ let g:gitgutter_sign_allow_clobber = 1
 
 #### Signs' colours and symbols
 
-By default vim-gitgutter uses your colourscheme's `Diff*` highlight groups' foreground colours for the signs' foreground colours.  For example, your `DiffAdd` foreground colour will be used for the `+` sign's foreground colour.
+The following rules are used by vim-gitgutter to choose the signs' colours:
+
+1. If your colourscheme defines the highlight groups `GitGutterAdd`, `GitGutterChange`, `GitGutterDelete`, etc., they will be used.
```
airblade: "If your vimrc or colorscheme..."

adah1972: No … as I mentioned in the discussion, defining the highlight groups in .vimrc does not work (for me).

adah1972: Re-testing shows defining the highlight groups in .vimrc works in Linux, but not in GUI MacVim (only). It might be a Mac-GUI-related bug.

adah1972:
@airblade Please check again whether the patch is satisfactory now. 😁

adah1972:
@airblade Is there anything else I should do?

airblade:
I believe most colour schemes devised within the past few years have good colours for `DiffAdd` etc and/or `GitGutterAdd`.  It's mainly the built-in ones from Vim's early days which don't – but in my opinion most people don't use those.

So the default option – using `GitGutterAdd` or `DiffAdd` – works well for the majority.

For those for whom the default colours don't work well, some kind of configuration is needed to change them.  Either we can introduce new global variables which activate some hard-coded colours (which the user might not like anyway); or we can explain how to change the colours using Vim's colour mechanism (including a snippet to copy-paste).  I think the latter is more useful because it's about Vim rather than a specific plugin.

adah1972:
> So the default option – using `GitGutterAdd` or `DiffAdd` – works well for the majority.

I am aghast on reading this. How can you say ‘majority’, when you do not have any numbers?

Frightening away novice users is never a good idea. In contrast, experienced Vim hackers can always find their ways around. That is why working out of the box is important!

airblade:
> How can you say ‘majority’, when you do not have any numbers?

That's why I said, "I believe..."

You haven't provided any numbers either.

I have always gone out of my way to make it work out of the box.  Please stop lecturing me.

Issue #688, opened 2020-03-30
lightline - hunk preview window status line strange behavior


qlem:
- latest commit SHA: 7c48aa1e726f73794fff99b4c03f9094565891d7
- vim version: 8.2.343

Hi,

### case 1
When I preview the current hunk through `:GitGutterPreviewHunk` the status line of the preview window looks like the default one (vim default status line) until I switch to an other window. Then the status line of the preview window is updated and looks as expected. I use lightline.

### case 2
If I run two times in a row `:GitGutterPreviewHunk`, the status line of the preview window looks as expected (lightline style) but she seems to have the focus while not. The cursor remains in the current window. I have to switch on the preview window and switch again to the previous window or another one to fix it.

airblade:
Assuming you're not using floating/popup windows for the preview, this is the code that opens the preview window:

https://github.com/airblade/vim-gitgutter/blob/7c48aa1e726f73794fff99b4c03f9094565891d7/autoload/gitgutter/hunk.vim#L438-L443

And this is the code that populates it:

https://github.com/airblade/vim-gitgutter/blob/7c48aa1e726f73794fff99b4c03f9094565891d7/autoload/gitgutter/hunk.vim#L487-L500

As you can see, it doesn't do anything in particular with the status line.

I'm afraid I have no idea why the status line is not looking the way you expect with lightline.

qlem:
Thank you for your answer, I will more investigate on this problem. 

However, the comment of the function `open_hunk_preview_window` line 396 say `Preview window: moves cursor to preview window.`. But in my case, the cursor doesn't move in the preview window once it is open. The cursor remains in the current window. And like i say in my 1st post, the status line of the preview window is updated as expected once I move the cursor on it.

Unfortunately i don't know well vim script. But is it possible that the problem is that the cursor doesn't move in the preview window once it is open ?

PS: I do not use floating windows option

airblade:
The cursor is supposed to remain in the current window; here's the code:

https://github.com/airblade/vim-gitgutter/blob/7c48aa1e726f73794fff99b4c03f9094565891d7/autoload/gitgutter/hunk.vim#L312

I'm not familiar with lightline but perhaps it only triggers and updates a window's status line when that window is focused.

What outcome would you like?  Simply that the preview's status line look lightline-y?

airblade:
You could try a map which calls `:GitGutterPreviewHunk` and then updates lightline.  For example (untested):

```viml
[...]

PR #689, opened 2020-03-31
fix the behavior of the status line of the hunk preview windows

qlem:
A fix to #688.

I removed `noautocmd` from the line that open the preview window and the line who moves back the cursor to the original window.

Like as said in the vim help about `noautocmd`: 

> To disable autocommands for just one command use the ":noautocmd" command modifier.  **This will set 'eventignore' to "all" for the duration of the following command.**

That is why the status line of the preview window is not updated as expected because the events are not emitted and so plugins like lightline cannot update the status line because they can't handle the relevant events.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -435,7 +435,8 @@ function! s:open_hunk_preview_window()
 
   silent! wincmd P
   if !&previewwindow
-    execute g:gitgutter_preview_win_location &previewheight 'new gitgutter://hunk-preview'
+    noautocmd execute g:gitgutter_preview_win_location &previewheight 'new gitgutter://hunk-preview'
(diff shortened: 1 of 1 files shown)


airblade:
The `noautocmd` is there to avoid the inefficiency of running code (triggered by events like `BufEnter`) designed for when the user switches to a new window.

In this case we just briefly flick in and out of the preview window; we don't stay there.

I wonder if there's a different way to allow the preview status line to be updated.

airblade:
Perhaps keep the `noautocmd` but then immediately fire `WinEnter`?

qlem:
Ok so I leaved the `noautocmd` and I added `doautocmd WinEnter gitgutter://hunk-preview` to fire `WinEnter` once after the opening of the preview window.

airblade:
Great, thanks!

PR #690, opened 2020-04-01
fix #688

### LINKED ISSUE #688: lightline - hunk preview window status line strange behavior

qlem:
fix issue #688

Sorry, I forgot to trigger `WinEnter` once the switch from the window preview to the original window is done. It's ok now.

However, the opening of windows or switch between them seem to trigger lot of events which may be tracked or required by other plugins or vim itself. I don't know if the use of `noautocmd` before critical action is a good idea.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -436,7 +436,7 @@ function! s:open_hunk_preview_window()
   silent! wincmd P
   if !&previewwindow
     noautocmd execute g:gitgutter_preview_win_location &previewheight 'new gitgutter://hunk-preview'
-    doautocmd WinEnter gitgutter://hunk-preview
+    doautocmd WinEnter
     let s:winid = win_getid()
     set previewwindow
     setlocal filetype=diff buftype=acwrite bufhidden=delete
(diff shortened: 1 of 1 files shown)


airblade:
> However, the opening of windows or switch between them seem to trigger lot of events which may be tracked or required by other plugins or vim itself.

I want to avoid all of those events because they are unnecessary when simply populating the preview window, may cause undesirable side-effects, any may slow things down.  They are not required by Vim.

airblade:
Next time, please use a better summary for your commits.

PR #692, opened 2020-04-10
update readme hunk cycle

derekacosta:
should be Prev so that it actually wraps

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -547,7 +547,7 @@ function! GitGutterNextHunkCycle()
   silent! GitGutterNextHunk
   if line('.') == line
     1
-    GitGutterNextHunk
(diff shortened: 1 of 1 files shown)


airblade:
The `1` in the previous line takes you to line 1, i.e. wraps the cursor, then you need `GitGutterNextHunk`.

derekacosta:
Maybe its may setup then, but have you tried it yourself?

airblade:
Yes, I tried it at the time and I've just tried it again now.  It's correct as it is.

airblade:
It takes you to next hunk (downwards) in the file.  When you reach the last one, it takes you to the first hunk in the file.

PR #709, opened 2020-05-28
"No hunk in file" when no previous nor next hunk

Biggybi:
As I often just want to check if there's a hunk, I though that a different message should be displayed when there's no hunk in the file.

I couldn't find a simple way to check if hunks are present in the file, and I wonder if such a small change deserves 30 lines of code.

I had to modify both **gitgutter#hunk#prev_hunk** and **gitgutter#hunk#next_hunk** functions to check for previous or next hunk, and display the message accordingly.

For the check, I added their counterparts **gitgutter#hunk#prev_hunk_check** and **gitgutter#hunk#next_hunk_check** (using the same algorithm).

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -58,7 +58,7 @@ function! gitgutter#hunk#next_hunk(count) abort
       endif
     endfor
     if gitgutter#hunk#summary(bufnr) == [0,0,0]
-        call gitgutter#utility#warn('No hunk in file')
+      call gitgutter#utility#warn('No hunk in file')
     else
       call gitgutter#utility#warn('No more hunks')
     endif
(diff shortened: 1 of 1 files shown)


airblade:
> I couldn't find a simple way to check if hunks are present in the file

How about this?

```viml
" in your vimrc
function! GitStatus()
  let [a,m,r] = GitGutterGetHunkSummary()
  return [a,m,r] == [0,0,0] ? '' : '[changes]'
endfunction
set statusline+=%{GitStatus()}
```

Biggybi:
Not sure how I could miss it. Well, the implementation is much cleaner with it.

Biggybi:
I agree there's not _much_ benefit at all. My status line does show if changes are present.
But, sometimes, I actually navigate my hunks to commit them (separately or not) and I don't really pay attention to the status bar (or it might even be too small for the hint to appear).
I may go up and down several times, and when the time has come when there are no changes left. I just think it makes more sense to have a more explicit message.
Indeed, that's a tiny nearly useless something, but, hey, why not?
I don't mean to insist, please feel free to do whatever :)

airblade:
Original discussion about what should be echoed: #236.

PR #725, opened 2020-07-27
fix full-width characters to half-width characters

jiroshin:
# problem
Currently, in my vim environment, the following error occurs at startup. (encoding=utf-8)
```
my-terminal $ vim
Error detected while processing function gitgutter#highlight#define_signs[8]..<SNR>31_define_sign_text:
line    5:
E239: Invalid sign text: _‾
```

According to vim doc, text option of sign must fit within two display cells. It is considered that the above error is due to the fact that this constraint is not satisfied when `_` and the character string(`‾`) are combined, because `‾` is a full-width character.
```
	text={text}						*E239*
		Define the text that is displayed when there is no icon or the
		GUI is not being used.  Only printable characters are allowed
		and they must occupy one or two display cells.
```

# solution
Use half-width strings instead of full-width strings.
`‾`(Unicode: FFE3) → `¯`(Unicode: 00AF)

### DIFF
--- a/doc/gitgutter.txt
+++ b/doc/gitgutter.txt
@@ -446,7 +446,7 @@ Defaults:
     let g:gitgutter_sign_modified           = '~'
     let g:gitgutter_sign_removed            = '_'
     let g:gitgutter_sign_removed_first_line = '‾'
-    let g:gitgutter_sign_removed_above_and_below = '_‾'
+    let g:gitgutter_sign_removed_above_and_below = '_¯'
     let g:gitgutter_sign_modified_removed   = '~_'
 <
 You can use unicode characters but not images.  Signs must not take up more than
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -62,7 +62,7 @@ else
   call s:set('g:gitgutter_sign_removed_first_line', '_^')
 endif
 
-call s:set('g:gitgutter_sign_removed_above_and_below', '_‾')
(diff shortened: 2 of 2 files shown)


airblade:
That's odd: it doesn't cause any problems in my Vim (8.2 patches 1-607 macOS).

Also, the character is currently an overline (Unicode: 203E), not a fullwidth macron (Unicode: FFE3).  Are you actually seeing the latter?

airblade:
Could this depend on the font?  I use Inconsolata.

Also, `:echo strwidth('‾')` (the current "top line" character) gives `1`.  Surely this means its width is 1 display cell?

kyoh86:
`strwidth` will not returns display cell.
In my case, maybe it depends on `'ambiwidth'`.

```
:set ambidwidth?
ambiwidth=double
```

airblade:
Ah, my `'ambiwidth'` is `single`.  When I set it to `double` the overline takes up 2 display cells.

airblade:
Would you be happy with Unicode: 0305?  I find 00AF a little too narrow.

kyoh86:
`0305` is COMBINING overline.
So it will not be shown alone.

`ā` <- a(0x61) + ` (0x0305)`

airblade:
Ok, I see: `U+203E` is `U+0020` + `U+0305`.

jiroshin:
The Unicode I wrote in the PR was wrong.
`Unicode: FFE3` → `Unicode: 203E`
Thank you!

PR #727, opened 2020-07-29
Set GitGutter colors, but respect defaults if they exist

jsit:
Set GitGutter colors, but respect defaults if they exist. No need for `set_sign_backgrounds` (which didn't work anyway).

This plugin was checking for whether certain highlights already existed, and then tried to avoid setting them if they did. The `:h :highlight-default` option does this for you; by saying `highlight def GitGutterAdd` instead of just `highlight GitGutterAdd`, GitGutter will leave any user-defined highlighting alone. This way GitGutter provides "sane defaults" (using the values for DiffAdd and DiffDelete etc.), but won't override a user's colors.

This PR will allow you to change colorschemes while GitGutter is loaded without GitGutter "going blank" on its colors.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -320,18 +320,6 @@ If you or your colourscheme has defined `GitGutter*` highlight groups, the plugi
 
 If you want the background colours to match the sign column, but don't want to update the `GitGutter*` groups yourself, you can get the plugin to do it:
 
-```viml
-let g:gitgutter_set_sign_backgrounds = 1
-```
-
-If no `GitGutter*` highlight groups exist, the plugin will check the `Diff*` highlight groups.  If their foreground colours differ the plugin will use them; if not, these colours will be used:
-
-```viml
-highlight GitGutterAdd    guifg=#009900 ctermfg=2
-highlight GitGutterChange guifg=#bbbb00 ctermfg=3
-highlight GitGutterDelete guifg=#ff2222 ctermfg=1
-```
-
 To customise the symbols, add the following to your `~/.vimrc`:
 
 ```viml
--- a/autoload/gitgutter/highlight.vim
+++ b/autoload/gitgutter/highlight.vim
@@ -70,29 +70,21 @@ function! gitgutter#highlight#define_highlights() abort
   " Highlights used by the signs.
 
   " When they are invisible.
-  execute "highlight GitGutterAddInvisible    guifg=bg guibg=" . guibg . " ctermfg=" . ctermbg . " ctermbg=" . ctermbg
-  execute "highlight GitGutterChangeInvisible guifg=bg guibg=" . guibg . " ctermfg=" . ctermbg . " ctermbg=" . ctermbg
-  execute "highlight GitGutterDeleteInvisible guifg=bg guibg=" . guibg . " ctermfg=" . ctermbg . " ctermbg=" . ctermbg
+  execute "highlight def GitGutterAddInvisible    guifg=bg guibg=" . guibg . " ctermfg=" . ctermbg . " ctermbg=" . ctermbg
+  execute "highlight def GitGutterChangeInvisible guifg=bg guibg=" . guibg . " ctermfg=" . ctermbg . " ctermbg=" . ctermbg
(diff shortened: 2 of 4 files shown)


airblade:
Can you give an example of something which doesn't work with the current code but does with yours?

jsit:
In the current master branch:

Minimal .vimrc:

```vim
set nocompatible                                       
set runtimepath=~/.vim/bundle/vim-gitgutter,$VIMRUNTIME
filetype plugin on                                     
filetype indent on                                     
syntax on                                              
color pablo
```

1. Open a file and make some changes to it to make the Git Gutter appear:

Screen Shot 2020-07-30 at 12 25 53 PM


2. Call `:verbose hi GitGutterDelete`

```
GitGutterDelete xxx ctermfg=12 ctermbg=242 guifg=Blue guibg=Grey
        Last set from ~/.dotfiles/vim/bundle/vim-gitgutter/autoload/gitgutter/highlight.vim line 90
```

Call `:color pablo`, and notice the Git Gutter goes "dark":

Screen Shot 2020-07-30 at 12 26 31 PM

4. Call `:verbose hi GitGutterDelete`:

```
GitGutterDelete xxx cleared
```

---

In this PR, do the same thing:

1. Open a file and make some changes to it to make the Git Gutter appear:

Screen Shot 2020-07-30 at 12 27 00 PM


2. Call `:verbose hi GitGutterDelete`

```
GitGutterDelete xxx ctermfg=12 ctermbg=242 guifg=Blue guibg=Grey
        Last set from ~/.dotfiles/vim/bundle/vim-gitgutter/autoload/gitgutter/highlight.vim line 86
```

3. Call `:color pablo`, and notice the Git Gutter doesn't go "dark":

Screen Shot 2020-07-30 at 12 27 23 PM

4. Call `:verbose hi GitGutterDelete`:

```
GitGutterDelete xxx ctermfg=12 ctermbg=242 guifg=Blue guibg=Grey
        Last set from ~/.dotfiles/vim/bundle/vim-gitgutter/autoload/gitgutter/highlight.vim line 86
```

Notice `GitGutterDelete` hasn't been cleared.

---

On `master`, you can't change color schemes without clearing Git Gutter's colors. In this PR, you can, and your theme foreground `Diff` colors are used. The code as-is isn't doing what it's trying to do.

airblade:
> This plugin was checking for whether certain highlights already existed, and then tried to avoid setting them if they did. 

That's not quite right.  It was checking whether certain highlights existed, and then merging a background colour (the sign column's colour) into them.  I couldn't just set default highlights because then their background colours might not match the sign column's colour.

> The :h :highlight-default option does this for you; by saying highlight def GitGutterAdd instead of just highlight GitGutterAdd, GitGutter will leave any user-defined highlighting alone. 

I understand but `highlight default` doesn't let you set just the background on an existing highlight.

Anyway, I'll look at the code now...

airblade:
I can solve the "going dark" problem like this:

```diff
diff --git i/autoload/gitgutter/highlight.vim w/autoload/gitgutter/highlight.vim
index 1439b72..2b68f9e 100644
--- i/autoload/gitgutter/highlight.vim
+++ w/autoload/gitgutter/highlight.vim
@@ -77,7 +77,7 @@ function! gitgutter#highlight#define_highlights() abort
 
   " When they are visible.
   for type in ["Add", "Change", "Delete"]
-    if hlexists("GitGutter".type)
+    if hlexists("GitGutter".type) && s:get_foreground_colors('Diff'.type) != ['NONE', 'NONE']
       if g:gitgutter_set_sign_backgrounds
         execute "highlight GitGutter".type." guibg=".guibg." ctermbg=".ctermbg
       endif
```

With that modification, the only effective difference between your code and mine – as far as I can tell – is that yours doesn't allow existing GitGutter* highlights to have their backgrounds automatically adjusted to match the sign column.  Please correct me if I'm wrong.

The idea behind that is that if you want to define your own sign colours, you probably just care about the foreground colours.  You don't want to figure out the sign column's colour and then update your custom signs with it – and why should you if the plugin could do it for you.

jsit:
Hm, the current solution looks odd to me; specifically, it seems to set the background color even if the user doesn't want it.

My new PR #729 revises the logic here a bit; please let me know what you think over there.

Issue #612, opened 2019-05-15
Signs do not disappear after commit from another tmux-window


astier:
## General Information

> What is the latest commit SHA in your installed vim-gitgutter?

7eeea63

> What vim/nvim version are you on?

NVIM v0.3.5

## Problem

After modifying a file in nvim in a tmux-window the gitgutter-signs do show up. However, if I switch to another tmux-window add/commit the changes and then switch back to the window with the opened file the signs do not disappear. I have to run :GitGutter manually.

airblade:
It sounds like `FocusGained` is not being triggered:

https://github.com/airblade/vim-gitgutter/blob/7eeea63e62b1cc088a75c7a7c244fc774d82e5bb/plugin/gitgutter.vim#L221-L223

Have you tried the suggestions in the [README](https://github.com/airblade/vim-gitgutter#when-signs-dont-update-after-focusing-vim)?

astier:
I solved it by setting `set -g focus-events on` in my tmux.conf
Thank you for your help.

airblade:
Glad you solved it.  I updated the docs with your solution.

PR #736, opened 2020-09-10
[doc] metion the plugin vim-tmux-focus-events

astier:
By default tmux doesn't send focus-events. The result is that the
autocmd `FocusGained` isn't triggered. The docs already mention to set
`set -g focus-events on` in the tmux.conf. However, gitgutter still
wouldn't update the signs when I committed the changes outside of vim.
To get it working I had to install the following plugin: [vim-tmux-focus-events](https://github.com/tmux-plugins/vim-tmux-focus-events)
I think it would be great to mention this plugin in the docs so other
tmux-users are able to fix this issue quickly.

### DIFF
--- a/doc/gitgutter.txt
+++ b/doc/gitgutter.txt
@@ -491,6 +491,8 @@ If you use tmux, try this in your tmux.conf:
 >
     set -g focus-events on
 <
+If it still doesn't work try using the following plugin:
+https://github.com/tmux-plugins/vim-tmux-focus-events
 
 When this option is 0, the plugin force-updates the buffer on |BufEnter|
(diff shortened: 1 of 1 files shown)


airblade:
> The docs already mention to set `set -g focus-events on` in the tmux.conf.

That was you: https://github.com/airblade/vim-gitgutter/issues/612#issuecomment-492699437 :)

At the time that seemed to be sufficient.  What changed?

astier:
I was sure something didn't work but I can't reproduce it anymore. I removed the plugin from my settings now and will see if it happens again by accident.

PR #744, opened 2020-11-01
Show message like "hunk 4 of 11" after hunk jumping

obcat:
Hi Andy!
Thank you for useful plugin :cat:

I added the new feature that a message like "hunk 4 of 11" is shown after hunk jumping.
what do you think?

gitgutter-pull-req-1

gitgutter-pull-req-2 gif

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -55,8 +55,6 @@ function! gitgutter#hunk#next_hunk(count) abort
   endif
 
   let current_line = line('.')
-  let hunk_total = len(hunks)
-  let hunk_number = 1
   let hunk_count = 0
   for hunk in hunks
     if hunk[2] > current_line
@@ -64,12 +62,11 @@ function! gitgutter#hunk#next_hunk(count) abort
       if hunk_count == a:count
         execute 'normal!' hunk[2] . 'Gzv'
         if g:gitgutter_show_msg_on_hunk_jumping
-          redraw | echo printf('Hunk %d of %d', hunk_number, hunk_total)
+          redraw | echo printf('Hunk %d of %d', index(hunks, hunk) + 1, len(hunks))
         endif
         return
       endif
     endif
-    let hunk_number += 1
   endfor
   call gitgutter#utility#warn('No more hunks')
 endfunction
@@ -85,8 +82,6 @@ function! gitgutter#hunk#prev_hunk(count) abort
   endif
 
   let current_line = line('.')
-  let hunk_total = len(hunks)
-  let hunk_number = hunk_total
   let hunk_count = 0
   for hunk in reverse(copy(hunks))
     if hunk[2] < current_line
@@ -95,12 +90,11 @@ function! gitgutter#hunk#prev_hunk(count) abort
         let target = hunk[2] == 0 ? 1 : hunk[2]
         execute 'normal!' target . 'Gzv'
         if g:gitgutter_show_msg_on_hunk_jumping
-          redraw | echo printf('Hunk %d of %d', hunk_number, hunk_total)
+          redraw | echo printf('Hunk %d of %d', index(hunks, hunk) + 1, len(hunks))
         endif
         return
       endif
     endif
-    let hunk_number -= 1
   endfor
   call gitgutter#utility#warn('No previous hunks')
 endfunction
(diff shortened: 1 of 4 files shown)


autoload/gitgutter/hunk.vim
```diff
@@ -55,15 +55,19 @@ function! gitgutter#hunk#next_hunk(count) abort
   endif
 
   let current_line = line('.')
+  let hunk_total = len(hunks)
+  let hunk_number = 1
```
airblade, line 59: I don't think you need a new variable.  You can use `hunk_count` and add 1 when you echo the message.

autoload/gitgutter/hunk.vim
```diff
@@ -55,15 +55,19 @@ function! gitgutter#hunk#next_hunk(count) abort
...
   let hunk_count = 0
   for hunk in hunks
     if hunk[2] > current_line
       let hunk_count += 1
       if hunk_count == a:count
         execute 'normal!' hunk[2] . 'Gzv'
+        redraw | echo printf('hunk %d of %d', hunk_number, hunk_total)
```
airblade, line 66: Please could capitalise the first letter, i.e. "Hunk ..."

airblade:
Thank you.

Vim doesn't echo messages when you navigate with `]c` / `[c` and generally I like to keep as consistent with Vim's behaviour as possible.  However this does look useful so I'm happy to include it if you can make it possible for the user to control whether or not anything is echoed.  I'm also happy for it to be opt-out instead of opt-in.

autoload/gitgutter/hunk.vim
obcat: I don't think so. When you type `3]c`, the value of `hunk_count` should be `3`, regardless of the hunk position.

autoload/gitgutter/hunk.vim
obcat: Okay :cat:

obcat:
I'm glad you're considering this request positively 😸 

> However this does look useful so I'm happy to include it if you can make it possible for the user to control whether or not anything is echoed. I'm also happy for it to be opt-out instead of opt-in.

Okay, I'll do it. Please give me some time.

autoload/gitgutter/hunk.vim
airblade: I don't understand.  `3]c` jumps forward 3 hunks from wherever the cursor is.

obcat: Suppose that you have 11 hunks and you are on the fourth hunk, i.e. `Hunk 4 of 11`. `3]c` jumps forward 3 hunks as you say, but the message we want to see is `Hunk 7 of 11` instead of `Hunk 3 of 11`.

airblade: Oh, I see.  Thanks :)

obcat:
Fixed, could you please review it? 🐱

autoload/gitgutter/hunk.vim
obcat: I don't like that `hunk_total` and `hunk_number` are unnecessarily declared even for users who turn `g:gitgutter_show_msg_on_hunk_jump` option off, but I don't think it affects performance so much.

obcat: I could simply add

```vim
if g:gitgutter_show_msg_on_hunk_jumping
  redraw | echo printf('Hunk %d of %d', index(hunks, hunk) + 1, len(hunk))
endif
```

without adding those variables, but I don't think this is  clever either.
Do you have any other good ideas? 🤔

airblade: Actually I quite like that ;)

airblade (commented):
Please could you also mention that the message can be turned off, in the Hunks section of the readme?

autoload/gitgutter/hunk.vim
obcat: Oh, okay. Then I replace it with the code above ;)

obcat:
> I'm also happy for it to be opt-out instead of opt-in.

You want to add this feature by default, i.e. set the default value of `g:gitgutter_show_msg_on_hunk_jumping` to `1` ?
I'm not very good at English so I may have been mistaken 😖

obcat:
> Please could you also mention that the message can be turned off, in the Hunks section of the readme?

Oh, I forgot about readme. I'll do it.

obcat:
Assuming that you want to add this feature by default, I've fixed the code ;)

airblade:
Thank you!  I squashed the commits together and merged them in 987a333.

obcat:
Thank you for your kind review and merge! 💛

Issue #751, opened 2020-12-09
<Leader>hs error with pwsh on W10


Acelya-9028:
Hello,

I use Neovim and pwsh on W10 and when I use `<Leader>hs`, `<Leader>hu` or `<Leader>hp` gitgutter crash.

![gitgutter_error](https://user-images.githubusercontent.com/45818681/101698784-63d11500-3a7a-11eb-8b86-43007107af2b.png)

Here is my shell config:

```
set shell=pwsh shellpipe=\| shellxquote=
set shellcmdflag=-NoLogo\ -NoProfile\ -ExecutionPolicy\ RemoteSigned\ -Command
set shellredir=\|\ Out-File\ -Encoding\ UTF8
```

I know that it's a problem with powershell when a temp file is used instead pipe because I have a similar probkem with [fugitive](https://github.com/tpope/vim-fugitive/issues/1644).

fugitive works with `set noshelltemp`  but gitgutter not :s.
gitgutter works with cmd.exe, but not with powershell or pwsh, with or without `set noshelltemp`, error is the same...

```
NVIM v0.4.4
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe /DWIN32 /D_WINDOWS /W3 /MD /Zi /O2 /Ob1 /DNDEBUG -DMIN_LOG_LEVEL=3 /W3 -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32 -D_WIN32_WINNT=0x0600 -DINCLUDE_GENERATED_DECLARATIONS -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -IC:/projects/neovim/build/config -IC:/projects/neovim/src -IC:/projects/nvim-deps/usr/include -IC:/projects/neovim/build/src/nvim/auto -IC:/projects/neovim/build/include
Compilé par appveyor@APPVYR-WIN

Features: -acl +iconv +tui
See ":help feature-compile"

         fichier vimrc système : "$VIM\sysinit.vim"
               $VIM par défaut : "C:/Program Files/nvim/share/nvim"

Run :checkhealth for more info
```

```
Nom du système d’exploitation:      Microsoft Windows 10 Entreprise LTSC
[...]

PR #752, opened 2020-12-11
Force cmd.exe to be used by system() instead of pwsh or powershell on Windows.

Acelya-9028:
cf #751.

### DIFF
--- a/autoload/gitgutter/utility.vim
+++ b/autoload/gitgutter/utility.vim
@@ -176,15 +176,20 @@ endfunction
 
 function! s:use_known_shell() abort
   if has('unix') && &shell !=# 'sh'
-    let [s:shell, s:shellcmdflag, s:shellredir] = [&shell, &shellcmdflag, &shellredir]
+    let [s:shell, s:shellcmdflag, s:shellredir, s:shellpipe, s:shellquote, s:shellxquote] = [&shell, &shellcmdflag, &shellredir, &shellpipe, &shellquote, &shellxquote]
     let &shell = 'sh'
     set shellcmdflag=-c shellredir=>%s\ 2>&1
+  endif
+  if has('win32') && (&shell ==# 'pwsh' || &shell ==# 'powershell')
+    let [s:shell, s:shellcmdflag, s:shellredir, s:shellpipe, s:shellquote, s:shellxquote] = [&shell, &shellcmdflag, &shellredir, &shellpipe, &shellquote, &shellxquote]
+    let &shell = 'cmd.exe'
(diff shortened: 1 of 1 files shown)


autoload/gitgutter/utility.vim
```diff
@@ -176,15 +176,20 @@ endfunction
...
   if has('unix') && &shell !=# 'sh'
-    let [s:shell, s:shellcmdflag, s:shellredir] = [&shell, &shellcmdflag, &shellredir]
+    let [s:shell, s:shellcmdflag, s:shellredir, s:shellpipe, s:shellquote, s:shellxquote] = [&shell, &shellcmdflag, &shellredir, &shellpipe, &shellquote, &shellxquote]
     let &shell = 'sh'
     set shellcmdflag=-c shellredir=>%s\ 2>&1
   endif
+  if has('win32') && (&shell ==# 'pwsh' || &shell ==# 'powershell')
+    let [s:shell, s:shellcmdflag, s:shellredir, s:shellpipe, s:shellquote, s:shellxquote] = [&shell, &shellcmdflag, &shellredir, &shellpipe, &shellquote, &shellxquote]
+    let &shell = 'cmd.exe'
+    set shellcmdflag=/s\ /c shellredir=>%s\ 2>&1 shellpipe=>%s\ 2>&1 shellquote= shellxquote="
```
airblade, line 186: The `shellxquote` value appears to be a double-quotation mark.  Is that right?  I would have expected `'"'` or `"\""` in that case.

airblade:
Do you have any other `shell*` options set in your vimrc, e.g. `shellcmdflag` or `noshelltemp`, which could affect this?

Acelya-9028:
It's my config:

```
set shell=pwsh shellpipe=\| shellxquote=
set shellcmdflag=-NoLogo\ -NoProfile\ -ExecutionPolicy\ RemoteSigned\ -Command
set shellredir=\|\ Out-File\ -Encoding\ UTF8
``` 

but the official is [runtime/doc/options.txt#L5209](https://github.com/neovim/neovim/blob/fbbc8cb8b110ebf05f33dd4bce83a59b7a79d822/runtime/doc/options.txt#L5209).

autoload/gitgutter/utility.vim
Acelya-9028: Here is my default:
shellxquote

airblade: Ah, I guess option values set with `:set` don't need quoting.

airblade:
We don't need to worry about `noshelltemp` because `system()` ignores it.

It sounds like we're setting and restoring the appropriate options.

Acelya-9028:
Thanks!

PR #761, opened 2021-02-06
Prevent hunk preview using preview window from changing alternate file

### LINKED ISSUE #760: Hunk preview using preview window changes alternate file
**Problem**
After hunk preview using preview window (`let g:gitgutter_preview_win_floating = 0`), the alternate file of current window will be changed.

**To Reproduce**

1. Suppose that the alternate file of current window (`window A`) is `foo.txt`.
    ```vim
    echo expand('#')
    " => foo.txt
    ```

1. Preview a hunk changes with:
    ```vim
    GitGutterPreviewHunk
    ```
    The hunk preview is shown in newly created preview window. This is named `gitgutter://hunk-preview`. The cursor still in `window A`.

1. The alternate file of `window A` is changed to the hunk preview buffer.
    ```vim
    echo expand('#')
    " => gitgutter://hunk-preview
    ```

**Expected behavior**
The alternate file is not changed.

**Additional context**

I think this is caused by a ~bug (?)~ strange behavior of the `:new {file}` command (See https://github.com/vim/vim/issues/7782). This command is used at:

https://github.com/airblade/vim-gitgutter/blob/ff9d134f0c69e25d391138036051b2c5e6bac864/autoload/gitgutter/hunk.vim#L464

**Environment**

> What is the latest commit SHA in your installed vim-gitgutter?

https://github.com/airblade/vim-gitgutter/commit/ff9d134f0c69e25d391138036051b2c5e6bac864

> What vim/nvim version are you on?

Vim 8.2.2164

obcat:
Fix #760

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -461,7 +461,8 @@ function! s:open_hunk_preview_window()
   if &previewwindow
     file gitgutter://hunk-preview
   else
-    noautocmd execute g:gitgutter_preview_win_location &previewheight 'new gitgutter://hunk-preview'
+    noautocmd execute g:gitgutter_preview_win_location &previewheight 'new'
+    file gitgutter://hunk-preview
     doautocmd WinEnter
     set previewwindow
   endif


airblade:
Thanks!

obcat:
You're welcome ;) Thank you for merging!

PR #762, opened 2021-02-08
New GitGutterSignsColumnToggle Feature!

jmembreno4194:
I have added a new feature that enables you to toggle the signs column on/off. It made it annoying to have to copy text from vim and have the empty column also get copied. Hope you guys accept to merge the feature.. And if maybe its not in the correct format, I'd appreciate it if you'd accept and move it where ever its supposed to go. I honestly didn't know where these types of features could be placed in. Best of luck to you guys 👍

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -317,6 +317,12 @@ By default the sign column will appear when there are signs to show and disappea
 set signcolumn=yes
 ```
 
+Or to toggle it, simply map the following command to any key in your vimrc:
+
+```
+:GitGutterSignsColumnToggle
+```
+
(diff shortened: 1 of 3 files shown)


airblade:
Thanks for the suggestion.

However you can get this behaviour with one line in your vimrc/init.vim:

```vim
:command! TSCL execute 'set signcolumn='.(&signcolumn == 'yes' ? 'no' : 'yes')
```

So I don't think building it into the plugin is justified.

But thanks again :)

jmembreno4194:
Actually, I do believe it would be pretty helpful.. How about beginners? Why don't you have that command on your README? Not everyone is gonna come up with your solution... that's why you pre-build most commands. You don't have to accept my pull request.. but maybe consider adding a feature like that or including a command in your README. Thanks for your time 👍

airblade:
Toggling the sign column is a vim thing, not a gitgutter-specific one, so it doesn't make sense to me to have it in this plugin.  What if you're using another plugin with signs and want to turn off the sign column?  Having to execute a gitgutter command would be odd, especially if you're not using gitgutter by then.

GitGutter already has `:GitGutterToggle`.  If you already have `:set signcolumn=auto` then the toggle command will close/open the sign column.

jmembreno4194:
Oh ok! Never mind then... Thanks for your time 👍

PR #763, opened 2021-02-09
Fix when pwsh or powershell is used as shell with a parameter.

Acelya-9028:
Like `set shell=pwsh\ -NoLogo`.

Actually, only `pwsh` or `powershell` are tested so use them with a parameter crash. See https://github.com/airblade/vim-gitgutter/issues/751.


airblade:
Thanks!

PR #765, opened 2021-02-12
fix(hunk): summary can't update after preview

kevinhwang91:
I'm not familiar with the code about gitgutter, but it can fix the issue as the title said.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -334,6 +334,7 @@ function! s:preview(hunk_diff)
   call s:enable_staging_from_hunk_preview_window()
   if &previewwindow
     call s:goto_original_window()
+    call gitgutter#process_buffer(bufnr(''), 1)
   endif
 endfunction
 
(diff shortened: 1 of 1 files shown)


kevinhwang91:
I don't use vim, it maybe uses `callback` opt for `popup_create`

airblade:
Sorry, I don't understand what problem this fixes.

What is the behaviour you expect to see and what do you actually see?

How can I reproduce the problem?

kevinhwang91:
1. `:GitGutterNextHunk`
2. `:GitGutterPreviewHunk`
3. `:echo GitGutterGetHunkSummary()` outputs `[0 ,0, 0]`
4. close preview hunk, `:echo GitGutterGetHunkSummary()` still outputs `[0 ,0, 0]`

airblade:
The hunk summary is for the whole file.  If you don't stage or undo any hunks, the hunk summary shouldn't change.

What looks weird to me is that the summary is `[0, 0, 0]`, which means there are no changes, and yet you seem to be navigating between hunks.  Does your file actually have any changes or not?

kevinhwang91:
https://asciinema.org/a/391090
Test under Linux with vim 8.2

airblade:
Thanks for the screencast which helped me understand the problem.

Thanks too for the patch.  However I think it fixes the symptoms rather than the actual cause, which I believe 6019d43 fixes.

kevinhwang91:
nice:)

PR #781, opened 2021-04-22
Clean up warning messages

VincentCordobes:
IMO warning messages are cleaner without the plugin name in. What do you think ?

Screenshot 2021-04-22 at 11 40 31

vs

Screenshot 2021-04-22 at 11 39 50

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -260,7 +260,7 @@ function! s:hunk_op(op, ...)
     call gitgutter#diff#process_hunks(bufnr, gitgutter#hunk#hunks(bufnr))  " so the hunk summary is updated
 
     if empty(s:current_hunk())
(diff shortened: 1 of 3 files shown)


airblade:
I put the plugin name as a prefix because I find it annoying when I get a random message echoed and I have no idea where it came from.

But I agree with you: the message is cleaner without the prefix.  Hopefully the messages are all clear enough not to confuse people.

Many of the messages start with a lower-case letter ([example](https://github.com/airblade/vim-gitgutter/blob/9756e95bd596a303946a90f06f4efe51dcd57e87/plugin/gitgutter.vim#L75)) because they are expecting to follow the prefix.  Please could you give them all an upper-case first letter?

VincentCordobes:
I updated the PR :)

airblade:
Thanks!

PR #789, opened 2021-07-04
GitGutterQuickFix for current file.

caojoshua:
Appends the path to the current file to the `git diff` command when `g:gitgutter_use_location_list = 1`.

New behavior:
* loclist only contains hunks of current file
* if current file is not within the repo, the loclist is empty
* if current buffer does not have a file loaded, the loclist is empty

I personally think this change is sensible when using the loclist. My personal config is (its neovim lua, but should make sense)
```
util.set_normal_keymap('<leader>ghq', util.cmd('GitGutterQuickFix') .. util.cmd('copen'))
util.set_normal_keymap('<leader>ghl', util.lua_cmd('vim.g.gitgutter_use_location_list=1') ..
    util.cmd('GitGutterQuickFix') .. util.lua_cmd('vim.g.gitgutter_use_location_list=0') .. util.cmd('lopen'))
```
I have keymaps to view hunks for the entire project, or just the current file. I can use the loclist by setting `g:gitgutter_use_location_list=1`, and then resetting it back to 0 afterwards.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -179,7 +179,7 @@ endfunction
 " - this runs synchronously
 " - it ignores unsaved changes in buffers
 " - it does not change to the repo root
-function! gitgutter#quickfix()
+function! gitgutter#quickfix(current_file)
   let cmd = g:gitgutter_git_executable.' '.g:gitgutter_git_args.' rev-parse --show-cdup'
   let path_to_repo = get(systemlist(cmd), 0, '')
   if !empty(path_to_repo) && path_to_repo[-1:] != '/'
@@ -191,6 +191,9 @@ function! gitgutter#quickfix()
         \ ' diff --no-ext-diff --no-color -U0'.
         \ ' --src-prefix=a/'.path_to_repo.' --dst-prefix=b/'.path_to_repo.' '.
         \ g:gitgutter_diff_args. ' '. g:gitgutter_diff_base
+  if a:current_file
+    let cmd = cmd.' -- '.expand('%:p')
+  endif
   let diff = systemlist(cmd)
   let lnum = 0
   for line in diff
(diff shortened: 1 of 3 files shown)


airblade:
This makes sense if you want your location list to hold locations from the current buffer only, but that's a personal preference.  Therefore I'd want to add an opt-in option to switch on this behaviour.

autoload/gitgutter.vim
```diff
@@ -191,6 +191,9 @@ function! gitgutter#quickfix()
         \ ' diff --no-ext-diff --no-color -U0'.
         \ ' --src-prefix=a/'.path_to_repo.' --dst-prefix=b/'.path_to_repo.' '.
         \ g:gitgutter_diff_args. ' '. g:gitgutter_diff_base
+  if g:gitgutter_use_location_list
+    let cmd = cmd.expand(expand('%:p'))
```
airblade, line 195: Why two `expand()`s?

Also I think it would be more robust to prefix the path with `--`.

caojoshua: huh thats odd. probably a typo. 

+1 on `--`

caojoshua:
Created `:GitGutterQuickFixCurrentFile` command, and no longer breaks current `g:gitgutter_use_location_list` functionality.

plugin/gitgutter.vim
```diff
@@ -118,7 +118,8 @@ command! -bar GitGutterBufferDisable call gitgutter#buffer_disable()
 command! -bar GitGutterBufferEnable  call gitgutter#buffer_enable()
 command! -bar GitGutterBufferToggle  call gitgutter#buffer_toggle()
 
-command! -bar GitGutterQuickFix call gitgutter#quickfix()
+command! -bar GitGutterQuickFix call gitgutter#quickfix(v:false)
+command! -bar GitGutterQuickFixCurrentFile call gitgutter#quickfix(v:true)
```
airblade, line 122: Please could you use `0`/`1` instead of `v:false`/`v:true` for more backward compatibility?  They were introduced in 7.4.1154 which is quite old but I try to support 7.4 altogether.

airblade:
> Created `:GitGutterQuickFixCurrentFile` command

Looks good!

plugin/gitgutter.vim
caojoshua: done.

airblade:
Thank you.

PR #794, opened 2021-08-04
Make intra-highlights customizable

obcat:
Hi @airblade.

Currently, the user cannot customize the color of the intra-highlights, "GitGutterAddIntraLine" and "GitGutterDeleteIntraLine", because we call the ":highlight" commands without "default" argument (:help :highlight-default). So I suggest adding it. What do you think?

Thank you.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -283,6 +283,9 @@ You can customise:
 * How to handle non-gitgutter signs
 * The signs' colours and symbols
 * Line highlights
+* Line number highlights (only in Neovim 0.3.2 or higher)
+* The diff syntax colours used in the preview window
+* The intra-line diff highlights used in the preview window
 * Whether the diff is relative to the index (default) or working tree.
 * The base of the diff
 * Extra arguments for `git` when running `git diff`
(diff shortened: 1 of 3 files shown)


airblade:
That would be fine.

Please could you also update the docs?

- the help doc in the [highlight](https://github.com/airblade/vim-gitgutter/blob/7d9b7125ff2533f6c520573f54e91db2f13a44b0/doc/gitgutter.txt#L571) section
- the readme, in a new section after the [line number highlights](https://github.com/airblade/vim-gitgutter#line-number-highlights) section

obcat:
Thank you for the positive response!

> Please could you also update the docs?
> 
> * the help doc in the [highlight](https://github.com/airblade/vim-gitgutter/blob/7d9b7125ff2533f6c520573f54e91db2f13a44b0/doc/gitgutter.txt#L571) section
> * the readme, in a new section after the [line number highlights](https://github.com/airblade/vim-gitgutter#line-number-highlights) section

Sure, please give me some time.

obcat:
I've updated the docs.

I also added a way to customize the diff syntax colours as well as intra-line highlighting, because I feel it a bit unnatural that the latter is only documented.

airblade:
Thanks!

obcat:
Yey! Thanks for the quick review and merge!

PR #795, opened 2021-08-06
Add option to enable looparound hunk jumping

BramVerb:
The option `g:gitgutter_looparound_on_hunk_jumping` can be turned on to
wrap around to the first hunk of the file when pressing next on the last hunk of the file.
When going backwards, it wraps around to the last hunk of the file (when you are at the first hunk).

Reasoning for this change: Since most of the vim behavior wraps around the end of the file, I wanted to have that for hunk jumping as well.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -43,6 +43,15 @@ function! gitgutter#hunk#increment_lines_removed(bufnr, count) abort
   call gitgutter#utility#setbufvar(a:bufnr, 'summary', summary)
 endfunction
 
+function! gitgutter#hunk#go_to(line_number, hunk_index, total_hunks) abort
+  execute 'normal!' a:line_number . 'Gzv'
+  if g:gitgutter_show_msg_on_hunk_jumping
+    redraw | echo printf('Hunk %d of %d', a:hunk_index + 1, a:total_hunks)
+  endif
+  if gitgutter#hunk#is_preview_window_open()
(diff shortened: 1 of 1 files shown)


autoload/gitgutter/hunk.vim
```diff
@@ -91,18 +99,18 @@ function! gitgutter#hunk#prev_hunk(count) abort
       let hunk_count += 1
       if hunk_count == a:count
         let target = hunk[2] == 0 ? 1 : hunk[2]
```
BramVerb, line 101: When changing this I didn't quite understand why this is only necessary for forward jumping. 
Can this special case also happen when jumping tot the next hunk?

If so, i can update it to check it in the `gitgutter#hunk#go_to)` as well. I think some more overall simplification can then be done.

airblade:
> Since most of the vim behavior wraps around the end of the file, I wanted to have that for hunk jumping as well.

Although searching wraps around the end of the file by default (depends on `'wrapscan'`), `]c` / `[c` do not and I try to remain consistent with Vim.  I suggested a [vimrc-based implementation](https://github.com/airblade/vim-gitgutter/issues/260#issuecomment-220057036) a while ago instead.

I'm not 100% against it, perhaps 90%.  I'll need to think about it.

BramVerb:
> I'm not 100% against it, perhaps 90%. I'll need to think about it.

I missed that there was a vimrc-based implementation, thanks for mentioning it. I'll use that instead then.

PR #798, opened 2021-08-30
fix quickfix for combined diffs

TwoFinger:
This is a fix for an error I was getting when using GitGutterQuickFix on diffs with conflicts:

ksnip_20210830-130651

Apparently, these diffs take `--cc` to produce output in the "combined" format.
git-diff(1) mentions `-c` too, but I haven't seen it in the wild.

On another note, both the existing and my new `elseif` branches for double-quoted filenames seem to get the special chars wrong.
If the filename contains, say, a Tab, the status line in vim displays something like `asdf^Iasdf`, but GitGutterQuickFix will try to open `asdf\tasdf`, which doesn't exist.
I haven't fixed that.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -205,6 +205,10 @@ function! gitgutter#quickfix(current_file)
     elseif line =~ '^diff --git "'
       let [_, fnamel, _, fnamer] = split(line, '"')
       let fname = fnamel ==# fnamer ? fnamel : fnamel[2:]
+    elseif line =~ '^diff --cc [^"]'
+      let fname = line[10:]
+    elseif line =~ '^diff --cc "'
+      let [_, fname] = split(line, '"')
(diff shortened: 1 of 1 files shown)


airblade:
I've never come across `--cc` or the combined format.  Are you inspecting a file which was affected by a merge commit?

Does `let g:gitgutter_diff_args='-m'` stop the problem occurring in the first place?

airblade:
Regarding special characters in the filename, I'll track that in a new issue.

TwoFinger:
> I've never come across `--cc` or the combined format. Are you inspecting a file which was affected by a merge commit?
No, it happens whenever I have conflicts during a rebase. No merge commits were involved.

> Does `let g:gitgutter_diff_args='-m'` stop the problem occurring in the first place?
No, made no difference.

airblade:
Thanks!

PR #802, opened 2021-09-11
work around for neovim if | foo | endif issue

mckellygit:
Hi,
ty for awesome gitgutter.

We seem to hit E171: Missing :endif
when at :s/ in neovim
see neovim/neovim#8796
I am not sure how to fix in neovim, so this is a simple work-around.

thx,
-m


mckellygit:
Just an update, there is now a neovim PR that should fix this -
https://github.com/neovim/neovim/pull/15638

mckellygit:
Closing this PR as the neovim fix is merged as this is no longer needed.

PR #807, opened 2021-11-01
Add border option

philipGaudreau:
This is a small proposition that permits having borders.

Also, I don't know if it was only me, but there seemed to be a glitch while doing a diff preview near the bottom of the window.
 
The comparison can be seen in the 2 pictures. The first is before.

Screen Shot 2021-10-31 at 23 40 47

Screen Shot 2021-10-31 at 23 38 11
.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -433,11 +433,13 @@ function! s:open_hunk_preview_window()
       " Set default width and height for now.
       let s:winid = nvim_open_win(buf, v:false, {
             \ 'relative': 'cursor',
-            \ 'row': 1,
+            \ 'row': -2,
             \ 'col': 0,
             \ 'width': 42,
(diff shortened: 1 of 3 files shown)


philipGaudreau:
Hi,

Thank you for the warm welcome!

> perhaps we should allow people to configure a dictionary of styling options which we merge into the options we pass

That is a very nice idea.

> Would you be interested in elaborating on your patch to do that?

I must admit that I'm still a beginner in Vimscript, and that I'm more familiar with Neovim's features than Vim's, but I'm willing to check it out, it should not be very hard to figure out. I will come back to you when there is more development.

lcrockett:
I would propose to make most of the `nvim_open_win()` options user configurable. `border`, `relative`, `row`, `col`, `style`, are amongst the most common ones to be modified based on user's taste. Perhaps expose the whole `nvim_open_win()` function so user can specify their own block of options ?

PR #810, opened 2021-12-10
Always pass some group name to sign arguments

0mp:
E.g., passing an empty group name to the sign argument linehl (e.g.,
"linehl=") causes the following error:

```
Error detected while processing /usr/home/0mp/.vim/plugged/vim-gitgutter/plugin/gitgutter.vim[96]..function gitgutter#highlight#define_signs[10]..<SNR>56_define_sign_line_highlights:
line    9:
E1249: Group name missing for linehl
```

This patch makes the those errors go away by passing "None" when no
group name is needed.

### DIFF
--- a/autoload/gitgutter/highlight.vim
+++ b/autoload/gitgutter/highlight.vim
@@ -175,12 +175,12 @@ function! s:define_sign_line_highlights() abort
     sign define GitGutterLineRemovedAboveAndBelow  linehl=GitGutterDeleteLine
     sign define GitGutterLineModifiedRemoved       linehl=GitGutterChangeDeleteLine
   else
-    sign define GitGutterLineAdded                 linehl=
-    sign define GitGutterLineModified              linehl=
-    sign define GitGutterLineRemoved               linehl=
-    sign define GitGutterLineRemovedFirstLine      linehl=
-    sign define GitGutterLineRemovedAboveAndBelow  linehl=
-    sign define GitGutterLineModifiedRemoved       linehl=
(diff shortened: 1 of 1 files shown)


0mp:
Tested on `VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec  7 2021 09:46:41)` coming from the `vim-gtk3-8.2.3745` package on FreeBSD.

airblade:
Thanks for the patch.

This usage of `linehl` always used be valid.  It looks like [8.2.3743](https://github.com/vim/vim/releases/tag/v8.2.3743) made it an error, and [8.2.3747](https://github.com/vim/vim/releases/tag/v8.2.3747) modified it, and finally [8.2.3748](https://github.com/vim/vim/releases/tag/v8.2.3748) reverted it.  There's a discussion [here](https://groups.google.com/g/vim_dev/c/0I4bu3yNnfE).

So I believe (without trying it) that if you upgrade Vim to 8.2.3748 the error will go away.  Given that, I don't think it's necessary to apply the patch.

coryschwartz:
I was just about to submit this exact same patch.

Currently the vim shipped in freebsd ports is affected, even though I don't see it anywhere else. Sure enough, it's running 8.2.3745 just like 0mp says.

airblade:
Ah, I didn't realise this particular Vim version was shipped in a package/port.

I've merged this now.  Thanks for the patch!

0mp:
Thanks a lot!

Issue #841, opened 2022-08-23
Suggestion: Can autocmd be avoided?


sak96:
> What is the latest commit SHA in your installed vim-gitgutter?
52d006682314a353a2ea951a8e443067f1db7eb4
> What vim/nvim version are you on?
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Aug 21 2022 14:20:23)

There is autocmd which check if gitgutter should run. This tends to update `v:shell_error` which may cause issue like: vifm/vifm.vim#80.

Not sure if there is easy way to avoid it but it is worth check out.

airblade:
The autocommands aren't your problem, it's `v:shell_error` getting set.  If two plugins are both executing shell commands via a mechanism like `system()` or `:!` there's going to be a conflict over which writes `v:shell_error` last.

Gitgutter obviously runs git shell commands.  Git uses exit codes to indicate outcomes.  Vim automatically sets `v:shell_error` when running shell commands synchronously.  However most of the time gitgutter uses jobs, running asynchronously, where exit codes are reported to the job handler not `v:shell_error`.  The only time it runs synchronously is at startup, due to a Vim bug (#619), unless you are using an ancient Vim (you didn't specify your version above) or you have configured gitgutter not to run async.

However in the linked issue gitgutter was responding to a `ShellCmdPost` autocommand, so it wasn't at startup.

I don't understand why your gitgutter appears to be running synchronously.  Please could you set `let g:gitgutter_log=1`, reproduce the problem, and upload the log file here?  It'll be `gitgutter.log` in the directory where gitgutter is installed.

sak96:
```logs


==== start log session ====

  0.000008 function startify#open_buffers[13]..<SNR>98_open_buffer[12]..BufEnter Autocommands for "*"..function <SNR>41_on_bufenter[22]..gitgutter#process_buffer[14]..<SNR>97_setup_path[3]..gitgutter#utility#set_repo_path[20]..gitgutter#utility#system[1]:
  0.000008 cd /home/$USER/project/distributed/cs6.824-rs/map_reduce/src && git  ls-files -v --error-unmatch --full-name -z -- test.rs

[...]

PR #842, opened 2022-08-24
Fix `gitgutter#async#available` to check patch version properly.

sak96:
As per `help has-patch` the patch version check should use `.` instead of `-`.

```help
<							*has-patch*
3.  Beyond a certain version or at a certain version and including a specific
    patch.  The "patch-7.4.248" feature means that the Vim version is 7.5 or
    later, or it is version 7.4 and patch 248 was included.  Example: >
	:if has("patch-7.4.248")
```

Note it is used properly in places like:
https://github.com/airblade/vim-gitgutter/blob/f19b6203191d69de955d91467a5707959572119b/autoload/gitgutter.vim#L30

Fixes: #841

### DIFF
--- a/autoload/gitgutter/async.vim
+++ b/autoload/gitgutter/async.vim
@@ -1,8 +1,8 @@
 let s:available = has('nvim') || (
       \   has('job') && (
-      \     (has('patch-7-4-1826') && !has('gui_running')) ||
-      \     (has('patch-7-4-1850') &&  has('gui_running')) ||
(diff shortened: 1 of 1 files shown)


sak96:
cc: @airblade

sak96:
i have validate that the `.` works in vim 8.1 as well as 9.0 
but `-` only works on 8.1

airblade:
Thanks, but already done in f19b620.

PR #854, opened 2023-04-15
Link highlight of GitGutterChangeDeleteLineNr to GitGutterChangeLineNr

ppwwyyxx:
This matches the behavior of `GitGutterChangeDeleteLine` and `GitGutterChangeDelete`, i.e., the "ChangeDelete" version always links to "Change".

The assumption is that most users don't need to distinguish the two. Making this change is useful so that these user don't have to do this link by themselves.


airblade:
Thank you!  I agree.

Issue #834, opened 2022-05-03
GitGutterDiffOrig toggle


ds2606:
> What is the latest commit SHA in your installed vim-gitgutter?
2f35907
> What vim/nvim version are you on?
8.2

Repeated invocations of `:GitGutterDiffOrig` keep opening more splits. To close the split, I manually move to the diff'ed split and close the buffer. (`:only` doesn't work in the case that there is originally >1 split before opening the diff).

Can we add a `:GitGutterDiffToggle` that just closes the GitGutter diff buffer if it's already open?

airblade:
Vim's own `:+clo` should do the trick.

airblade:
The `+` means apply the `clo[se]` command to the next window.  It's a `{count}`.  See `:help edit-window` for possible values you can use.

If your `'splitright'` is off, you need to use `:-clo` instead or you'll get `E16: Invalid range`.

ds2606:
Nifty, just learned more about count specifiers, thanks!

ds2606:
The only kink is that I have `:GGDiffOrig` mapped to `<leader>gd`: it'd be nice just to be able to just invoke that same keybinding again to close the just-opened diff buffer. 

I'm going to just implement a short function to take care of this, but seems like something that could be handy for users in general

airblade:
> OK, so I just implemented this in a super basic way. I'm new to vimscript and not sure if this is hacky or not though

Looks good to me.

You could perhaps avoid using a variable to track the state by looking for a side-effect of the command.  For example:

```vim
function! GitGutterDiffOrigToggle()
  if &diff
    +clo
  else
    GitGutterDiffOrig
  endif
endfunction
```

You could collapse this further:

```vim
nnoremap <silent> <expr> <leader>gd &diff ? ':+clo<CR>' : ':GitGutterDiffOrig<CR>'
```

ds2606:
Much better/cleaner, thanks :-)

PR #857, opened 2023-04-25
Added :GitGutterDiffOrigToggle and made :GitGutterDiffOrig would not …

drazil100:
…open duplicate difforig buffers

One feature I felt was lacking in gitgutter is a way to quickly pull up :GitGutterDIffOrig and then dismiss it just as quickly.

### How this works:
There is now a dictionary named s:difforigbuffers keyed to the buffer number of the buffer the diff is for with the value set to the number of the buffer that gets created. Whenever :GitGutterDiffOrigToggle or `gitgutter#difforig_toggle()` gets called it first loops through the `s:difforigbuffers`dictionary and checks if the current buffer number already exists as a key or value in the dictionary. If it is a value or it is a key that currently has a value that is a listed buffer then it closes the buffer and returns. If it isn't a key or isn't a value or the difforig buffer isn't listed then it calls gitgutter#difforig()

I also modified gitgutter#difforig() to check the same dictionary to prevent duplicate buffers from being created.

Let me know if there is anything that needs to be changes or if there are any weird edge cases I am not accounting for. Also I didn't know where to put the dictionary so I just put it above gitgutter#difforig()

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -237,7 +237,16 @@ function! gitgutter#quickfix(current_file)
 endfunction
 
 
+let s:difforigbuffers = {}
+
 function! gitgutter#difforig()
+  for [k, v] in items(s:difforigbuffers)
+    if bufnr('%') == k && buflisted(v) || bufnr('%') == v
+      " This either already has an open difforig buffer or is a difforig
+      " buffer. Do nothing and return.
+      return
+    endif
+  endfor
   let bufnr = bufnr('')
   let path = gitgutter#utility#repo_path(bufnr, 1)
   let filetype = &filetype
@@ -262,4 +271,22 @@ function! gitgutter#difforig()
   diffthis
   wincmd p
   diffthis
+  let s:difforigbuffers[bufnr('%')] = bufnr('$')
+endfunction
+
+
+function! gitgutter#difforig_toggle()
+  for [k, v] in items(s:difforigbuffers)
+    if bufnr('%') == k && buflisted(v) || bufnr('%') == v
+      " This either already has a difforig buffer, or this is a difforig
+      " buffer. Close this file's difforig buffer and wipe it from the buffer
+      " list.
+      execute 'bwipeout ' . v
+      let s:difforigbuffers[k] = ''
+      return
+    endif
+  endfor
+
+  " Open a new difforig for the current buffer.
+  call gitgutter#difforig()
 endfunction
--- a/plugin/gitgutter.vim
+++ b/plugin/gitgutter.vim
@@ -135,6 +135,7 @@ command! -bar GitGutterQuickFix call gitgutter#quickfix(0)
 command! -bar GitGutterQuickFixCurrentFile call gitgutter#quickfix(1)
 
 command! -bar GitGutterDiffOrig call gitgutter#difforig()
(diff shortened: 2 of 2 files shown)


drazil100:
I forgot to mention. The reason I created this is because I wanted to be able to map it to a key for a quick look back at what was changed without having to switch splits to close the difforig buffer. This is also safer as it makes it less likely you will close the actual file and leave only the difforig buffer behind.

airblade:
Thanks for the suggestion.

This came up about a year ago.  Does the solution [here](https://github.com/airblade/vim-gitgutter/issues/834#issuecomment-1118436863) work for you?

drazil100:
I always forget to check the closed issues xD

Just gave it a try and it isn't quite as clean in functionality. With that solution it throws an error if you are not in a diff split
```
Error detected while processing function ToggleGitGutterDiffOrig:
line    2:
E16: Invalid range:     +clo
```
What I did works regardless of if you are in the diff split or the actual file's split as it actually tracks the buffer numbers with no split changing required. My solution does work on it's own without having to be part of the plugin and if you would prefer I keep it to my personal config I completely understand, but I would be happy to make any changes you want to make this work in your plugin if you are interested. I realize now in my excitement that I forgot to edit the documentation (though you may prefer to be the one to do that)

airblade:
I thought about this some more and remembered that I implemented `:GitGutterDiffOrig` to match Vim's `:DiffOrig`.  I want to keep it aligned with `:DiffOrig` so it behaves as people would expect.

Which means that I'll say no, thank you, to your patch.

Thanks very much anyway!

drazil100:
In that case I'll just leave this here for anyone reading this later who wishes to replicate my toggle setup in their personal 
vim / nvim configs. Thanks for taking the time to consider this feature.

lua:
```lua
-- Table of difforig buffers and buffers that spawned them
local my_diff_buffers = {}

function ToggleGitGutterDiffOrig()
  for k, v in pairs(my_diff_buffers) do
    if fn.bufnr('%') == k and vim.api.nvim_buf_is_valid(v) or vim.fn.bufnr('%') == v then
      -- The buffer is open, so close the split
      vim.api.nvim_buf_delete(v, { force = true })
      my_diff_buffers[k] = nil
      return
    end
  end
  -- The buffer is closed, so open the split and store the buffer number
  cmd("GitGutterDiffOrig")
  my_diff_buffers[vim.fn.bufnr('%')] = vim.fn.bufnr('$')
end

vim.api.nvim_set_keymap('n', '<F2>', '<cmd>lua ToggleGitGutterDiffOrig()<CR>', {silent=true})
```

vimscript:
```vim
" Table of difforg buffers and the buffers that spawned them
let s:my_diff_buffers = {}

function! ToggleGitGutterDiffOrig()
  for [k, v] in items(s:my_diff_buffers)
    if (bufnr('%') == k && buflisted(v)) || bufnr('%') == v
      " The buffer is open, so close the split
      execute 'bwipeout ' . v
      call remove(s:my_diff_buffers, k)
      return
    endif
  endfor

  " The buffer is closed, so open the split and store the buffer number
  call gitgutter#difforig()
  let s:my_diff_buffers[bufnr('%')] = bufnr('$')
endfunction

nnoremap <silent> <F2> :call ToggleGitGutterDiffOrig()<CR>
```

PR #862, opened 2023-06-14
fix(diff): Wrong diff if dos format combines with noeol

tcx4c70:
If fileformat is dos and eol is not set, then a '\r' (but without '\n') will be written to the last line of temporary buffer file, which will confuse git and make git think all lines are modified. Fix it by adding 'r' to all lines except the last one if fileformat is dos and eol is not set.

### DIFF
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -407,7 +407,13 @@ function! s:write_buffer(bufnr, file)
   endif
 
   if getbufvar(a:bufnr, '&fileformat') ==# 'dos'
-    call map(bufcontents, 'v:val."\r"')
+    if getbufvar(a:bufnr, '&endofline')
+      call map(bufcontents, 'v:val."\r"')
+    else
+      for i in range(len(bufcontents) - 1)
+        let bufcontents[i] = bufcontents[i] . "\r"
+      endfor
+    endif
   endif
 
   if getbufvar(a:bufnr, '&endofline')


airblade:
Thanks for reporting this.

Your solution looks good.  Before I can merge it, please could you post a simple file, e.g.:

```
foo
bar
```

– in dos format with eol not set, so I can try this out myself and write a test?

airblade:
Thanks for the file, it's useful.

Your patch doesn't do what you want it to do.  This line:

```vim
call map(bufcontents[0:-2], 'v:val."\r"')
```

...does not update all bar the last item of `bufcontents`.  `bufcontents[0:-2]` produces a new list so updating it does not update the original `bufcontents`.

The neatest way I could think of is to update every item but then undo the last item's update when `'eol'` is off:

```diff
diff --git i/autoload/gitgutter/diff.vim w/autoload/gitgutter/diff.vim
index a3edb12..b3e4bfc 100644
--- i/autoload/gitgutter/diff.vim
+++ w/autoload/gitgutter/diff.vim
@@ -408,6 +408,9 @@ function! s:write_buffer(bufnr, file)
 
   if getbufvar(a:bufnr, '&fileformat') ==# 'dos'
     call map(bufcontents, 'v:val."\r"')
+    if !getbufvar(a:bufnr, '&endofline')
+      let bufcontents[-1] = bufcontents[-1][:-2]
+    endif
   endif
 
   if getbufvar(a:bufnr, '&endofline')
```

Alternatively we could do a for loop over `bufcontents` from the first to the penultimate item, appending `"\r"`.  Hmm, that actually feels better because we're not making an incorrect change and then backing it out.

airblade:
```diff
diff --git i/autoload/gitgutter/diff.vim w/autoload/gitgutter/diff.vim
index a3edb12..9e016a4 100644
--- i/autoload/gitgutter/diff.vim
+++ w/autoload/gitgutter/diff.vim
@@ -407,7 +407,13 @@ function! s:write_buffer(bufnr, file)
   endif
 
   if getbufvar(a:bufnr, '&fileformat') ==# 'dos'
-    call map(bufcontents, 'v:val."\r"')
+    if getbufvar(a:bufnr, '&endofline')
+      call map(bufcontents, 'v:val."\r"')
+    else
+      for i in range(len(bufcontents) - 1)
+        let bufcontents[i] = bufcontents[i] . "\r"
+      endfor
+    endif
   endif
 
   if getbufvar(a:bufnr, '&endofline')
```

tcx4c70:
Thanks for your reviewing and suggestion!

During my test, I find another problem on Windows platform. If the original fileformat is dos and you change it to unix via `:set fileformat=unix` and `:w`, then gitgutter won't show any changes (it should show that all lines of the file have been changed).

tcx4c70:
> During my test, I find another problem on Windows platform. If the original fileformat is dos and you change it to unix via :set fileformat=unix and :w, then gitgutter won't show any changes (it should show that all lines of the file have been changed).

Never mind, it should be a problem of my git config (`core.autocrlf = true`) and everything is OK if I set `core.autocrlf` to `false` on Windows.

airblade:
Thanks for confirming my patch works for you.

PR #871, opened 2023-08-31
Remove "file unknown in base" check

### LINKED ISSUE #869: vim-gitgutter clobbers v:shell_error via autocmd ShellCmdPost
> What is the latest commit SHA in your installed vim-gitgutter?

5f525476f3ade40e60aa295a8a2a01c0935b0d61

> What vim/nvim version are you on?

    $ vim --version
    VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Aug 27 2023 17:04:14)
    macOS version - arm64
    Included patches: 1-1800
    Compiled by Homebrew
    [...]

### Setup

1. Open any file tracked by git that did not exist in `g:gitgutter_diff_base`, e.g. `let g:gitgutter_diff_base = 'HEAD^'` and open a file that was added in `HEAD`
2. `:!echo foobar`
3. `:echo v:shell_error`

### Expected result

`:echo v:shell_error` should print `0`

### Actual result

`:echo v:shell_error` prints `128`

## Problem

Due to `autocmd ShellCmdPost * call gitgutter#all(1)` in [`plugin/gitgutter.vim`](https://github.com/airblade/vim-gitgutter/blob/5f525476f3ade40e60aa295a8a2a01c0935b0d61/plugin/gitgutter.vim#L323), `gitgutter#diff#run_diff` is called immediately after `:!echo foobar`, which [calls `gitgutter#utility#system(cmd)`](https://github.com/airblade/vim-gitgutter/blob/5f525476f3ade40e60aa295a8a2a01c0935b0d61/autoload/gitgutter/diff.vim#L92) in order to determine `"file unknown in base"`. This sets `v:shell_error` to `128`, and we lose `v:shell_error` from the `:!echo foobar` that we just ran.

The opposite is similarly problematic: if we're editing a file that _does_ exist in `g:gitgutter_diff_base`, and we run something like `:!exit 1`, then `:echo v:shell_error` after that prints `0`, not `1`.

I stumbled across #841 while figuring this out, but the issue here is different; I confirmed that `:echo gitgutter#async#available()` and `:echo g:gitgutter_async` both print `1`.

## Workarounds (outside of vim-gitgutter)

1. From [vifm/vifm.vim#80](https://github.com/vifm/vifm.vim/issues/80) I see that one workaround is to use `:noautocmd` for any individual `:!cmd`, and I've confirmed that works for simple cases like this, but 1) it might not be desirable to temporarily disable all autocmds like that, and 2) this places the burden on other plugins and user code to handle this case.

2. It's possible to entirely disable this autocmd via `autocmd! gitgutter ShellCmdPost *`, though this doesn't work in vimrc, and must be in e.g. a `~/.vim/after/plugin/disable-gitgutter-ShellCmdPost.vim`. Same downsides as the other workaround.

## Possible solutions (in vim-gitgutter)

1. Assuming `g:gitgutter_async && gitgutter#async#available()`, then similar to how the actual diff is run via [`gitgutter#async#execute`](https://github.com/airblade/vim-gitgutter/blob/5f525476f3ade40e60aa295a8a2a01c0935b0d61/autoload/gitgutter/diff.vim#L172), would it be possible to check for `"file unknown in base"` async too? (I'm not super familiar with the code here so this might be nonsensical or impossible, but I figure it's worth asking)

2. Conceptually similar to how `winsaveview()` + `winrestview()` works, can we manually restore the previous `v:shell_error`?

nkouevda:
Instead of checking whether this file exists in `diff_base` and returning early, we let `git show` fail, which still creates an empty `from_file` for the subsequent `git diff` to use.

This mostly reverts 2ee95686c5944f99b42dd04fec005b30497006de, but we keep the test case.

For posterity: this was originally meant to fix #869, but #870 accomplished that.

### DIFF
--- a/autoload/gitgutter.vim
+++ b/autoload/gitgutter.vim
@@ -46,8 +46,6 @@ function! gitgutter#process_buffer(bufnr, force) abort
         call gitgutter#debug#log('Not tracked: '.gitgutter#utility#file(a:bufnr))
       catch /gitgutter assume unchanged/
         call gitgutter#debug#log('Assume unchanged: '.gitgutter#utility#file(a:bufnr))
-      catch /gitgutter file unknown in base/
-        let diff = gitgutter#diff#hunk_header_showing_every_line_added(a:bufnr)
       catch /gitgutter diff failed/
         call gitgutter#debug#log('Diff failed: '.gitgutter#utility#file(a:bufnr))
         call gitgutter#hunk#reset(a:bufnr)
--- a/autoload/gitgutter/diff.vim
+++ b/autoload/gitgutter/diff.vim
@@ -81,20 +81,6 @@ function! gitgutter#diff#run_diff(bufnr, from, preserve_full_diff) abort
     throw 'gitgutter assume unchanged'
   endif
 
-  " If we are diffing against a specific branch/commit, handle the case
-  " where a file exists on the current branch but not in/at the diff base.
-  " We have to handle it here because the approach below (using git-show)
-  " doesn't work for this case.
-  if !empty(g:gitgutter_diff_base)
-    let index_name = gitgutter#utility#get_diff_base(a:bufnr).':'.gitgutter#utility#repo_path(a:bufnr, 1)
-    let cmd = gitgutter#git().' --no-pager show '.index_name
-    let cmd = gitgutter#utility#cd_cmd(a:bufnr, cmd)
-    let [_, error_code] = gitgutter#utility#system(cmd)
-    if error_code
-      throw 'gitgutter file unknown in base'
-    endif
-  endif
-
   " Wrap compound commands in parentheses to make Windows happy.
   " bash doesn't mind the parentheses.
   let cmd = '('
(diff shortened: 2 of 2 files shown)


airblade:
I didn't expect this to work, given what I wrote [here](https://github.com/airblade/vim-gitgutter/issues/869#issuecomment-1702359568) about the error handler being triggered, but it does indeed work – nice!  It turns out the output on stderr from the failing `git-show` comes before the diff output on stdout, so although the error callback is triggered it doesn't matter because the success callback is triggered afterwards.

My only concern is `;`.  It works fine for me on macOS.  But according to this [Windows command shell overview](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490954(v=technet.10)#using-multiple-commands-and-conditional-processing-symbols) on Windows the semi-colon is used differently, to separate parameters.  It looks like the correct character is `&`.  However that would (probably?) need escaping, which is a pandora's box.

So I'd really like to merge this but I can't unless I can be sure it won't break on Windows.

airblade:
> However that would (probably?) need escaping...

Having said that, the existing `&&` and `||` in the command aren't escaped so maybe it's ok.  Do you have any way to try it on Windows?

nkouevda:
Rebased and updated with a different approach: adding another `|| exit 0`.

So instead of:

```bash
cd ... && (git show ... && git diff ... | grep ... || exit 0)
```

It would now be:

```bash
cd ... && (git show ... || exit 0) && (git diff ... | grep ... || exit 0)
```

> Do you have any way to try it on Windows?

I do not.

airblade:
Thanks, this is much neater than my original solution.

PR #884, opened 2024-02-08
Display git status in tabline

nbelakovski:
This is some work I did on my own and I thought that including it in gitgutter made the most sense. Here's a screenshot of what this looks like (with the dracula colorscheme)

image

I think it's pretty self explanatory. If you'd like to include this, don't merge it just yet, let me know and I can modify it to use gitgutter highlight groups. 

If, on the other hand, you don't want to include this, please let me know so I can get started on releasing it on my own.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -9,6 +9,7 @@ The name "gitgutter" comes from the Sublime Text 3 plugin which inspired this in
 Features:
 
 * Shows signs for added, modified, and removed lines.
+* Shows git status in the tabline
 * Runs the diffs asynchronously where possible.
 * Ensures signs are always up to date.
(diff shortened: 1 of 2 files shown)


airblade:
Thanks for posting this!

I'd like to keep gitgutter as small as I can and I'm all for people building their own extensions on top of it.  For example, gitgutter doesn't do anything with the status line, but instead makes it [easy for info to be incorporated into the status line](https://github.com/airblade/vim-gitgutter?tab=readme-ov-file#status-line) if people want to.

So I think you should go ahead and release your code as its own plugin.

nbelakovski:
OK, thanks for the quick action!

PR #890, opened 2024-04-28
Respect &foldopen during next/prev hunk jumps

lukelbd:
This makes the next/previous hunk block-style mappings respect the `&foldopen` setting by skipping `zv` if it does not contain the word `'block'`.

Think this would be more consistent with other vim behavior, and since `&foldopen` contains `'block'` by default it shouldn't affect the average user. However I can see the argument that this is a "breaking" change requiring an explicit setting like `g:gitgutter_respect_foldopen` (happy to update the PR if you feel this way).

Thanks for the awesome plugin!

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -60,7 +60,8 @@ function! gitgutter#hunk#next_hunk(count) abort
     if hunk[2] > current_line
       let hunk_count += 1
       if hunk_count == a:count
-        execute 'normal!' hunk[2] . 'Gzv'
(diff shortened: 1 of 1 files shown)


airblade:
This is great, thanks!  Being as consistent as possible with other vim behaviour is always the goal.

 Isn't `'foldopen'` a global option?  And therefore can't we just inspect `&foldopen` instead of `&l:foldopen`?

I don't think we need to introduce a new setting for this.

lukelbd:
Great! And yep you're right it's global (didn't check) -- just force pushed a version with `&foldopen`

airblade:
Brilliant, thanks again!

PR #903, opened 2025-03-05
Move to floating window upon retriggering preview

tohalla:
Moving to floating windows is sometimes quite frustrating, especially with multiple windows (the simplest case would be with a single split and attempting to navigate to the floating window via `<C-w>`. Some plugins have adopted a pattern to move to the floating window upon retriggering the command to open the float when the window is already open.

### DIFF
--- a/autoload/gitgutter/hunk.vim
+++ b/autoload/gitgutter/hunk.vim
@@ -359,6 +359,11 @@ endfunction
 
 
 function! s:preview(hunk_diff)
+  if g:gitgutter_preview_win_floating && exists('*nvim_set_current_win') && s:winid != 0
+    call nvim_set_current_win(s:winid)
+    return
+  endif
+
   let lines = split(a:hunk_diff, '\r\?\n')
   let header = lines[0:4]
   let body = lines[5:]


airblade:
Thanks for the suggestion.  I hardly ever use floating windows myself so I don't have a feel for the ergonomics.

Can't you just use `<C-w><C-w>` or `<C-w> w` to move into a floating window?

Also, does "some plugins" mean one or two, or is there more of a consensus than that?

tohalla:
I'm personally aware of:
* https://github.com/prabirshrestha/vim-lsp/blob/04428c920002ac7cfacbecacb070a8af57b455d0/autoload/vital/_lsp/VS/Vim/Window/FloatingWindow.vim#L217
* https://github.com/rhysd/git-messenger.vim

`<C-w>` would be fine otherwise, but having multiple windows open at same time makes it a pain to track where one should navigate, plus in case you have the same buffer open on other window, the floating window is closed due to
```
    autocmd CursorMoved,TabLeave <buffer> ++once call gitgutter#hunk#close_hunk_preview_window()
```
`<C-W> P` works fine for preview window, which I was using previously due to the navigation issue.

airblade:
Thank you!

PR #916, opened 2026-04-06
Fix/buffilepost throw concatenation

germtb:
#  Fix missing string concatenation in s:on_buffilepost throw

The throw on line 269 of plugin/gitgutter.vim is missing . between the string literal and the variables. This causes E488 whenever the error path is hit, which happens when plugins like fzf.vim trigger multiple rapid BufFilePre/BufFilePost events.

I added a regression test.


airblade:
Thanks!

I think I had `echo` originally but then changed it to `throw`, forgetting that they handle their arguments differently.

